Issue #26, opened 2020-07-28
Use inconspicuous loader


rgaudin:
While we now have a *transparent* chrome, we still display the webrecorder logo at loading time. We'd like something less colorful and more generic.

Issue #23, opened 2020-07-28
Prevent dead links


rgaudin:
Following-up on #21, we want to create ZIM files without dead links. This is slightly different to WARC which focuses on fidelity with the original source.

Preventing dead links means:
* not including files with no content (0 bytes length)
* rewriting content to exclude links to missing articles.
* keeping links to external URLs

ikreymer:
Yes, sometimes we also call this the 'web archive boundary' problem: what happens when a user hits the edge.
Is the desired behavior to replicate the behavior of `/catch/external`?
What should happen if a link is a 404 that is not external? (could be some dynamically created url)

rgaudin:
@kelson42 should be able to answer.

For wikis, we remove links to same-wiki, non included articles and leave foreign links as is (could be blocked or not at runtime)

In zimit, the scope is fuzzy as one can scrape only a subset of a website (or multiples) 
so it's hard to tell.

We could say that for warc2zim, the WARC is our context and whatever is not included is a remote link, even if on same domain or visually a _next of kin_

kelson42:
I believe I still miss a proper understanding of what is `/catch/external`... but I believe as well as we don't want/need any kind of SW based catching of external URLs (beside the one which can be specified in kiwix-serve).

The most important is that there is no internal dead links within the ZIM files (this is what zimcheck try to detect). Every URL which is not an internal link is an external link, ie. starting with a protocol (like `https://` for example) and which should just respect the online original.

ikreymer:
> I believe I still miss a proper understanding of what is `/catch/external`... but I believe as well as we don't want/need any kind of SW based catching of external URLs (beside the one which can be specified in kiwix-serve).
> 
> The most important is that there is no internal dead links within the ZIM files (this is what zimcheck try to detect). Every URL which is not an internal link is an external link, ie. starting with a protocol (like `https://` for example) and which should just respect the online original.

The full urls from the web page are always preserved, so there maybe be many links that start with `https://`, they are rewritten in the service worker to be loaded internally.. The way that this would work is that the service worker can detect if a URL is not found, and render a proper response, which could be: an error page, a redirect confirmation message, or automatically redirect. It can also do different things based on url, filering by domain.. I'll see what can be done..

ikreymer:
I think this should be resolved now as much as is possible. The redirect handling should be working now.
Of course, it is impossible to prevent things that are missing on the live site.. Bouquineux has some PDFs that are 0-length/blank for example

[...]

Issue #22, opened 2020-07-28
External link to Wikipedia in Bouquineux seems broken


kelson42:
Loading `bouquineux.zim` in a local kiwix-serve I open the following page:
http://localhost:8080/bouquineux/A/viewer.html#url=http%3A%2F%2Fwww.bouquineux.com%2F%3Febooks%3D346%26N%25C3%25A9mirovsky&ts=20200727182643

On that page there is a link to Wikipedi article:
http://localhost:8080/bouquineux/A/wabac/id-3d79eb55abd2/20200727182643mp_/http://fr.wikipedia.org/wiki/Ir%C3%A8ne_N%C3%A9mirovsky

This URL returns a HTTP 404:
```
$ curl -I http://localhost:8080/bouquineux/A/wabac/id-3d79eb55abd2/20200727182643mp_/http://fr.wikipedia.org/wiki/Ir%C3%A8ne_N%C3%A9mirovsky
HTTP/1.1 404 Not Found
Connection: Keep-Alive
Content-Length: 2344
Cache-Control: no-cache, no-store, must-revalidate
Access-Control-Allow-Origin: *
Content-Type: text/html
Date: Tue, 28 Jul 2020 09:20:32 GMT

```

and it should route to the online Wikipedia article https://fr.wikipedia.org/wiki/Ir%C3%A8ne_N%C3%A9mirovsky like the original page.

Remark: I believe we would really benefit here to keep the original link, ie. `https://fr.wikipedia.org/wiki/Ir%C3%A8ne_N%C3%A9mirovsky` in place of `/A/wabac/id-3d79eb55abd2/20200727182643mp_/http://fr.wikipedia.org/wiki/Ir%C3%A8ne_N%C3%A9mirovsky`

rgaudin:
Might be related to #23

ikreymer:
Yes, I just found the `/catch/external` that shows the warning page , I think that can easily be supported. Assume that is what would be desired here?

kelson42:
@ikreymer Not really I believe, what is the best is just keeping the link like it was.

rgaudin:
> Yes, I just found the `/catch/external` that shows the warning page , I think that can easily be supported. Assume that is what would be desired here?

No we want external links to stay as is ; unless blocked by kiwix-serve (optional)

Issue #21, opened 2020-07-28
Zimcheck does not pass with bouquineux.zim


kelson42:
```
$zimcheck  bouquineux.zim 
[INFO] Checking zim file bouquineux.zim
[INFO] Verifying Internal Checksum.. 
  [INFO] Internal checksum found correct
[INFO] Searching for metadata entries..
[INFO] Searching for Favicon..
[INFO] Searching for main page..
[INFO] Verifying Articles' content.. 
[INFO] Searching for redundant articles..
  Verifying Similar Articles for redundancies..
[ERROR] Missing metadata entries :
  Entry A/www.bouquineux.com/pdf/Tchekhov-La_Steppe.pdf is empty
  Entry A/www.bouquineux.com/pdf/Tchekhov-Trois_ans.pdf is empty
  Entry A/www.bouquineux.com/pdf/Tolstoi-Contes_et_nouvelles_-_Tome_III.pdf is empty
[ERROR] Missing metadata entries :
  Title
  Creator
  Publisher
  Description
[ERROR] Missing favicon :
_Map_base::at
```

kelson42:
I guess #4 already handles the lack of metadata/favicon... but the missing PDF seems to be an other problem.

rgaudin:
There are three problems listed here:

* Articles without data (see #23)
* Missing metadata (see #1 and #4)
* Missing favicon entry at `-/favicon` (see #4)

@ikreymer, if you don't have it, you can get zimcheck in the [zim-tools](http://download.openzim.org/release/zim-tools/).

ikreymer:
These files are empty from the site:
http://www.bouquineux.com/pdf/Tolstoi-Contes_et_nouvelles_-_Tome_III.pdf
http://www.bouquineux.com/pdf/Tchekhov-Trois_ans.pdf
http://www.bouquineux.com/pdf/Tchekhov-La_Steppe.pdf

I don't think empty responses should be considered errors, this is what the site offers.
If they are excluded, then it will just be a 404/not found, which seems less correct for the user..

ikreymer:
A solution to fix validation would be to write the `H/...` record but not the `A/...` record, and if the headers indicate that the payload is 0 length, don't attempt to load it at all.

ikreymer:
After removing the empty entries, `zimcheck` still seems to be confused by some URLs that are actually in the ZIM.
[...]

PR #28, opened 2020-07-29
Multi-Page Mode via SW

ikreymer:
This sets up a different mode, as suggested in #20, where the rendering of the original urls is taken over by the service workers.
This uses the low-level https://github.com/webrecorder/wabac.js directly instead of replayweb.page and that adds some flexibility.

Since URLs are preserved, suggestions, search and random should work as in the old system.
This system adds several static files that are stored in the `./replay/` folder and also installs sw.js directly from npm.
Indexing and language are also enabled with this. Also, empty payload records are not written to address #21 

The new files in `./replay` are:
- `index.html` - initial page, which will redirect to the main page. Also displays loading message or error message.
- `load.js` - the script used in `index.html`
- `topFrame.html` - the top frame page used on all pages. The banner is injected here by kiwix-serve.
- `sw.js` - the service worker, installed from npm automatically. Loaded from: https://cdn.jsdelivr.net/npm/@webrecorder/wabac/
- `notFoundPage.html` - Generic error page template for when a 404 is encountered. Redirects to live web if not in current domain. Also checks if kiwix-serve is blocking redirects and then redirects to `/catch/external`

This PR addresses:
 - #5 - the wabac.js is bundled directly
 - #17 - random should work
 - #18 - suggestions should work
 - #19 - search should work
 - #20 - no more single page
 - #21 - don't write empty payload A/ records, only H/ records.
 - #22 and #23 - external links should now redirect. if the blocking message is enabled in kiwix-serve it will be displayed.
 - #26 - new loading is just a text message, can be modified in `index.html`


src/warc2zim/replay/index.html
```diff
@@ -0,0 +1,11 @@
+<html>
+<head>
```
rgaudin, line 2: You need to set encoding otherwise the taskbar might be displayed incorrectly if ZIM name uses non-ASCII chars.

rgaudin: Notfound and Top as well

src/warc2zim/replay/notFoundPage.html
```diff
@@ -0,0 +1,26 @@
...
+<script>
+  const URL = "$URL";
+  const mainUrl = "$MAIN_URL";
+
+  if (window.parent === window.top) {
+
+    // check if URL is from same domain as the main url
+    if (!URL.startsWith(mainUrl)) {
+      // check if kiwix-serve 'blocking' is enabled, show interstitional if it is
+      if (window.block_path) {
```
rgaudin, line 12: Please submit a PR on kiwix-lib to add a comment on top of this so maintainers know that other tools rely on this
https://github.com/kiwix/kiwix-lib/blob/5f8c0998291d0778887373355bb6b1c5a345ed2c/static/skin/block_external.js#L1

src/warc2zim/replay/topFrame.html
```diff
@@ -0,0 +1,42 @@
...
+  height: 100%;
+  margin: 0;
+  padding: 0;
+  border: 0;
+  overflow: hidden;
+}
+
+iframe {
+  width: 100%;
+  height: 100%;
```
rgaudin, line 16: Might be a viewport problem but this revived the unreachable content bug ; in FF this time.

src/warc2zim/replay/topFrame.html
```diff
@@ -0,0 +1,42 @@
...
+  window.addEventListener("message", function() {
+    if (event.data.wb_type === "load" || event.data.wb_type === "replace-url") {
+      
+      // remove scheme to be consistent with current canonicalization
+      const urlNoScheme = event.data.url.slice(event.data.url.indexOf("//") + 2);
+      window.history.replaceState(null, "", PREFIX + urlNoScheme);
+    }
+  });
+
+  iframe.src = PREFIX + "mp_/" + URL;
```
rgaudin, line 39: Is `mp_` a sort of namespace?  Where does this come from? sw.js?

src/warc2zim/main.py
```diff
@@ -21,6 +21,8 @@
 from argparse import ArgumentParser, RawTextHelpFormatter
```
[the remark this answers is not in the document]
ikreymer: Ok, will remove in subsequent PR

src/warc2zim/main.py
```diff
@@ -155,55 +163,33 @@ def get_data(self):
...
+class RWPStaticArticle(BaseArticle):
     def __init__(self, filename, main_url):
-        super(RWPViewerArticle, self).__init__()
+        super(RWPStaticArticle, self).__init__()
         self.filename = filename
         self.main_url = main_url
 
+        self.mime, _ = mimetypes.guess_type(filename)
+        self.content = pkg_resources.resource_string(
+            "warc2zim", "replay/" + filename
```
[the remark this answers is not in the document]
ikreymer: This is using `pkg_resources` to load data from the package, which always use `/`, not the file system:
https://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access

There may be better ways to do this, will look.

src/warc2zim/main.py
```diff
@@ -224,21 +210,28 @@ def __init__(self, args):
...
-            )
+        self.add_remote_or_local("sw.js")
+
+        for filename in pkg_resources.resource_listdir("warc2zim", "replay"):
+            if filename != "sw.js":
+                self.replay_articles.append(RWPStaticArticle(filename, self.main_url))
 
         with Creator(
-            self.name, main_page="viewer.html", index_language="", min_chunk_size=8192
+            self.name, main_page="index.html", index_language="en"
```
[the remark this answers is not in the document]
ikreymer: This is fixed in another branch that I'll add after..

src/warc2zim/replay/index.html
ikreymer: Good point! Will fix in follow up.

src/warc2zim/replay/topFrame.html
ikreymer: Yes, sort of, it is a part of the replay.. Probably needs better documentation in wabac.js
I have it documented in pywb, which is the python version of the rewriting system, but it is the same idea:
https://pywb.readthedocs.io/en/latest/manual/rewriter.html#main-page-modifier-mp

Without the modifier, it serves the outer frame.
With the `mp_/` modifier it serves the rewritten content.

Issue #33, opened 2020-08-03
Not Found page should not be text/HTML


rgaudin:
notFoundPage.html being added as `text/html`, it receives a taskbar from kiwix-serve which results in a double taskbar.

Issue #32, opened 2020-08-03
Set encoding on HTML files


rgaudin:
Lack of encoding in HTML files can break taskbar display

Issue #31, opened 2020-08-03
Suggestions should display title instead of URL


rgaudin:
Article's title is currently set to URL which makes suggestions offer URLs instead of titles.

Depends on #29

rgaudin:
Maybe a duplicate of #27. Not sure what is squeezed as I don't have any.

Issue #30, opened 2020-08-03
Title and favicon from source should be displayed


rgaudin:
Main frame should set the correct source title and favicon.

See #29.

kelson42:
If the title has been fixed, it seems the favicon is not properly available.

See this screenshot. On the left kiwix-serve, on the right the original web site.

![image](https://user-images.githubusercontent.com/1029718/97287774-4424bb00-1845-11eb-8260-fff5a13f55d7.png)

PR #34, opened 2020-08-03
Support for User-Specified and Auto-Detected/Extracted Metadata

ikreymer:
Add cmdline options for metadata (title, description, creator, publisher, source, and tags)
Add automatic metadata for scraper and flavour
Auto extract title for all pages
Auto-detect main url (if not specified), title, favicon, and language from main url.
Add option to include all urls (-a), or urls from specified domains only, or urls only from domain of main url (the default).
Improvements to HTML templates, including adding charset=UTF-8
Favicon and Title passed from iframe to top frame as well.

Fixes:
- #1 - support for tags
- #4 - other metadata
- #6 - main url detection
- #29 - title and language parsing
- #27 - favicon fixes
- #31 - suggestions use titles
- #32 - sets encoding on HTML templates
- #30 - favicon and title from replay frame set on topFrame
- #33 - not found page now served as `application/octet-stream` from kiwix-serve by using .dat extension
- #35 - indexing only HTML pages
- #40 - try downloading favicon if doesn't exist in warc


src/warc2zim/replay/topFrame.html
```diff
@@ -29,10 +34,31 @@
...
+      if (event.data.title) {
+        document.title = event.data.title;
+      }
       
       // remove scheme to be consistent with current canonicalization
       const urlNoScheme = event.data.url.slice(event.data.url.indexOf("//") + 2);
       window.history.replaceState(null, "", PREFIX + urlNoScheme);
+
+      // if icons received, replace any existing icons with new ones
+      if (event.data.icons) {
```
rgaudin, line 47: does hat mean that browsing twitter.com will bring twitter favicon but browsing to somthing.net (with no favicon) will keep twitter favicon?

tests/test_warc_to_zim.py
```diff
@@ -11,17 +11,21 @@
 from warc2zim.main import warc2zim
 
 
-WARCS = [
-    "example-response.warc",
-    "example-resource.warc.gz",
-    "example-revisit.warc.gz",
-    "example-utf8.warc",
-    "netpreserve-twitter.warc",
```
rgaudin, line 19: can you explain why you are removing a test file? usually not a good sign 😉

ikreymer: Oh, just to save space mostly :) The single-page-test.warc has one page from lesfondamentaux with favicon, so thought it would be a better test. It can also be used with the --include-domain option to include only a single domain. 

The other WARC has pages from multiple sites but no favicon, which should still work of course, but was less a primary use case..
I suppose I could keep it if you'd like.. It would need to be used with the `-a` flag to include all pages from all domains in the WARC.

src/warc2zim/replay/topFrame.html
ikreymer: good question! I don't remember that happening in other environment, but will double check!

tests/test_warc_to_zim.py
rgaudin: That's fine. As you didn't mention it, I wanted to make sure we're not hiding problematic tests under the carpet :)

src/warc2zim/main.py
```diff
@@ -90,39 +98,30 @@ def get_data(self):
...
 # ============================================================================
-class WARCPayloadArticle(BaseWARCArticle):
+class WARCPayloadArticle(BaseArticle):
     """ WARCPayloadArticle used to store the WARC payload
     Usually stored under A namespace
     """
 
     def __init__(self, record):
-        super(WARCPayloadArticle, self).__init__(record)
+        super(WARCPayloadArticle, self).__init__()
```
rgaudin, line 107: you can do `super().__init__()`

src/warc2zim/main.py
```diff
@@ -47,35 +55,35 @@ def get_filename(self):
...
-# ============================================================================
-class WARCHeadersArticle(BaseWARCArticle):
+class WARCHeadersArticle(BaseArticle):
     """ WARCHeadersArticle used to store the WARC + HTTP headers as text
     Usually stored under H namespace
     """
 
     def __init__(self, record):
-        super(WARCHeadersArticle, self).__init__(record)
+        super(WARCHeadersArticle, self).__init__()
```
rgaudin, line 77: `super().__init__()`

src/warc2zim/main.py
```diff
@@ -90,39 +98,30 @@ def get_data(self):
...
-            # if the record has HTTP headers, use the Content-Type from those (eg. 'response' record)
-            return self.record.http_headers["Content-Type"]
-
-        # otherwise, use the Content-Type from WARC headers
-        return self.record.rec_headers["Content-Type"]
+        # TODO: converting text/html to application/octet-stream to avoid rewriting by kiwix
+        # original mime type still preserved in the headers block
+        if self.mime == "text/html":
+            self.mime = "application/octet-stream"
+            self.title = parse_title(self.payload) or self.url
```
rgaudin, line 118: I think we should fallback to `""`. That's how we do elsewhere. If source page has no title, then it's not findable via suggest.

src/warc2zim/main.py
```diff
@@ -198,14 +214,38 @@ def __init__(self, args):
...
+
+        self.metadata = {
+            "name": args.name,
+            "description": args.desc,
+            "creator": args.creator,
+            "publisher": args.publisher,
+            "tags": ";".join(args.tags) or None,
+            # optional
+            "source": args.source,
+            "flavour": os.path.basename(sys.argv[0]) + " " + " ".join(sys.argv[1:]),
```
rgaudin, line 246: There is no Flavour for warc2zim files. You should just not set it.

src/warc2zim/main.py
```diff
@@ -219,64 +259,201 @@ def add_remote_or_local(self, filename):
...
+            if self.main_url != url:
+                continue
+
+            if mime not in HTML_TYPES:
+                logger.warning(
+                    "Main page is not an HTML Page, mime type is: {0} - Skipping Favicon and Language detection".format(
+                        mime
+                    )
+                )
+
```
rgaudin, line 326: didn't you miss a `continue` here?

src/warc2zim/main.py
```diff
@@ -219,64 +259,201 @@ def add_remote_or_local(self, filename):
...
+def get_record_mime_type(record):
+    if record.http_headers:
+        # if the record has HTTP headers, use the Content-Type from those (eg. 'response' record)
+        content_type = record.http_headers["Content-Type"]
+    else:
+        # otherwise, use the Content-Type from WARC headers
+        content_type = record.rec_headers["Content-Type"]
+
+    mime = content_type or ""
+    return content_type.split(";")[0]
```
rgaudin, line 447: not using `mime`?

src/warc2zim/main.py
```diff
@@ -309,12 +485,35 @@ def warc2zim(args=None):
...
+    )
+
+    parser.add_argument("-f", "--favicon", help="Favicon for Main Page")
+
+    # optional metadata
+    parser.add_argument("--name", help="The name of the ZIM", default="")
+    parser.add_argument("--title", help="The Title", default="")
+    parser.add_argument("--desc", help="The Description", default="")
+    parser.add_argument("--tags", action="append", help="One or more tags", default=[])
+    parser.add_argument("--lang", help="Language", default="")
```
rgaudin, line 513: Specify that it should be ISO-639-3

src/warc2zim/replay/topFrame.html
ikreymer: To confirm, yes, if there is no icon, this will be an empty list, and `if([])` if true in JS, so it will just remove any existing icons

src/warc2zim/main.py
ikreymer: Fixed, was a typo!

src/warc2zim/main.py
ikreymer: Added some clarification.. The warning is to indicate the main page url is not an html page, it'll still be allowed (i guess could be image or pdf or something) but is not technically an error..

src/warc2zim/main.py
ikreymer: Added more info.

src/warc2zim/main.py
ikreymer: set to ""

src/warc2zim/main.py
```diff
@@ -219,64 +259,204 @@ def add_remote_or_local(self, filename):
...
+                return
+
+        if record.rec_type != "revisit":
+            yield WARCHeadersArticle(record)
+            payload_article = WARCPayloadArticle(record)
 
-                    if record.rec_type != "revisit":
-                        zimcreator.add_article(WARCHeadersArticle(record))
-                        payload_article = WARCPayloadArticle(record)
+            if len(payload_article.payload) != 0:
```
rgaudin, line 428: Shouldn't you check the headers before creating the PayloadArticle first? So you can skip early if Content-Length is null. You could still test of the header is not set and that would still save this test for the 99.9999% of the records where we do have a payload.

src/warc2zim/main.py
```diff
@@ -294,11 +474,10 @@ def warc2zim(args=None):
...
 
     parser.add_argument(
-        "-n",
-        "--name",
-        help="""Base name for WARC file, appropriate extension will be
-                                added automatically.""",
-        metavar="name",
+        "-o",
+        "--output",
+        help="""Output filename for ZIM file (.zim extension will be added)""",
```
rgaudin, line 479: When setting output explicitly, we should let them add the extension.

src/warc2zim/main.py
```diff
@@ -309,12 +488,39 @@ def warc2zim(args=None):
...
+    )
+
+    parser.add_argument(
+        "-i",
+        "--include-domains",
+        action="append",
+        help="List of domains that should be included. Not used if --include-all is set. Defaults to domain of the main url",
+    )
+
+    parser.add_argument("-f", "--favicon", help="Favicon for Main Page")
```
rgaudin, line 509: Not clear about what's expected here.

src/warc2zim/main.py
```diff
@@ -309,12 +488,39 @@ def warc2zim(args=None):
...
+
+    # optional metadata
+    parser.add_argument("--name", help="The name of the ZIM", default="")
```
rgaudin, line 512: This should be required with no default

src/warc2zim/main.py
```diff
@@ -309,12 +488,39 @@ def warc2zim(args=None):
...
+    parser.add_argument("--title", help="The Title", default="")
+    parser.add_argument("--desc", help="The Description", default="")
```
rgaudin, line 514: Can we use `description` so it's alike other scrapers?

src/warc2zim/main.py
```diff
@@ -309,12 +488,39 @@ def warc2zim(args=None):
...
+    parser.add_argument("--tags", action="append", help="One or more tags", default=[])
+    parser.add_argument(
+        "--lang",
+        help="Language (should be a ISO-639-3 language code). If unspecified, will attempt to detect from main page, or use 'eng'",
+        default="",
+    )
+    parser.add_argument("--publisher", help="ZIM publisher", default="-")
```
rgaudin, line 521: Could default to `Kiwix`, as in other scrapers

src/warc2zim/main.py
```diff
@@ -198,14 +214,38 @@ def __init__(self, args):
             logger.setLevel(logging.INFO)
 
         self.indexed_urls = set({})
-        self.name = args.name
-        if not self.name:
-            self.name, ext = os.path.splitext(args.inputs[0])
-            self.name += ".zim"
+        self.output = args.output
+        if not self.output:
```
rgaudin, line 218: As stated in #4, if no output is specified, it should be derived from `--name` and include the period (YYYY-MM).
What we do with other scrapers is that we have 3 args:
- `--output` (defaults to `/output`)
- `--name` mandatory
- `--zim-file` just a filename in output folder. built from `--name` if not specified

ikreymer:
> Not part of the PR but creating a temp dir and moving into it in tests seems a little odd. Wouldn't it be clearer to just use the pytest-provided `tmp_dir` fixture?

The moving into temp dir is to support the behavior that the automatic name of the zim is the same file and directory as the WARC. So if input is `/some/path/filename.warc` the default output is `/some/path/filename.zim`. To avoid creating files in the data directory, the `/some/path/filename.warc` is moved so its `<tmpdir>/filename.warc`.

But perhaps don't need that behavior? If so, can remove.. I see above that you suggested `/output` should be the default output dir, so that would remove the need for this copy.

src/warc2zim/main.py
ikreymer: ok, this can replace the current behavior, which uses the name of the WARC file as the name of the ZIM file by default.

src/warc2zim/main.py
ikreymer: There could be a `Transfer-Encoding: chunked` on the HTTP headers which means we don't know what the payload is until we read it fully.. Though there are ways to compute it by subtracting the current stream position from total length of record, I don't think that's even worth it here. Since we're always reading the payload if it exists, it's not really much of an optimization even then.. I guess would just save a constructor call maybe? Seems a very very minor difference, and there shouldn't be many of these anyway... Could also raise an exception in constructor if payload is 0 but again, seems like its making it more complicated than what is there now : )

src/warc2zim/main.py
ikreymer: Update help text and will now attempt to download the URL if not present in warc.

src/warc2zim/main.py
rgaudin: OK, that's fine then

Issue #36, opened 2020-08-04
i18n user-facing strings


rgaudin:
At least the not-found page present English text strings to user. We should have i18n for those and at least one translation stub (french). Instructions on how to set-it up is visible on /openzim/youtube repo.

ikreymer:
I think this should also be handled by kiwix-lib, ideally, as it is part of the replay system.
For the service worker error, #43 suggests a way that this would be done via kiwix-lib.
For not found page, how does this work currently? Is there a localized version of the not found page in kiwix-lib now?
If so, probably should find a way to just load the existing error text to be consistent.

rgaudin:
I think there's a misunderstanding about what kiwixlib is for. First kiwixlib, as its name suggests is for kiwix tools to use. Zim readers don't need kiwixlib, they only need libzim or any implementation of the spec.
Putting anything in kiwixlib implies it's not available for non-kiwix readers and translating strings from a scraper should be in the scraper. If you have the strings in your scraper repo, you're responsible for setting up the translation.

Second, we don't modify kiwixlib for every scraper. What we're doing with warc2zim is different enough from other scraper that we'll have to add changes to the architecture and potentially the Zim spec. kiwixlib should be content agnostic.

Also, we're talking about a very limited set of strings here ; and I provided a link on how to implement that easily. 
Should this eventually move elsewhere, it would be very easy to remove from here but until then, this should be internationalized.

ikreymer:
The youtube scraper has [text specific to youtube](https://github.com/openzim/youtube/blob/master/youtube2zim/locale/fr/LC_MESSAGES/messages.po) (playlist, etc...)

> At least the not-found page present English text strings to user.

Unless I'm misunderstanding something, shouldn't it ideally just be the same as this page?
https://github.com/kiwix/kiwix-lib/blob/master/static/templates/404.html

How does this page get internationalized, or it doesn't currently?

Just want to make sure I understand - can copy that file into every ZIM and add i18n.

rgaudin:
> The youtube scraper has [text specific to youtube](https://github.com/openzim/youtube/blob/master/youtube2zim/locale/fr/LC_MESSAGES/messages.po) (playlist, etc...)

Absolutely and this scraper has text specific to this scraper (not found string, service worker not working string, loading string – that is being replaced by visual one though, maybe others?). It's not much but it's English-only at the moment so should get i18n'd.

> Unless I'm misunderstanding something, shouldn't it ideally just be the same as this page?
[...]

PR #44, opened 2020-08-07
localization: setup localization using jinja2 and babel (#36)

ikreymer:
- setup localization of all templates (rename replay -> templates)
- add sample localized strings for fr
- run all templates (except sw.js) through jinja
- store string tables in package data
- babel supported via setuptools
- using new 'text/html;raw=true' mime for text/html to avoid rewriting


ikreymer:
@rgaudin is this what you had in mind for localization? will add brief docs also, using babel with setuptools integration

rgaudin:
@ikreymer yep that's it. Didn't know the babel/setuptools integration, love it! 
I've updated the code to make use of scraperlib ; check latest commit and fixed the french strings.

ikreymer:
Yes, my translations were with google translate :)

Though, some possible issues with the change:
- The default on command line is unset so that it can be detected from html. If it defaults to "eng", then it will not use the HTML detection. (This is why it was `fr` as it was testing extraction from lesfondamentaux page)
- I didn't use zimscraperlib because loading the .mo file from pkg_data, so its not tied to the directory structure. This should allow `warc2zim` to be run from any directory after being installed via pip -- I think zimscraperlib expects the `locale` directory to exist in current working directory.

rgaudin:
>     * The default on command line is unset so that it can be detected from html. If it defaults to "eng", then it will not use the HTML detection. (This is why it was `fr` as it was testing extraction from lesfondamentaux page)

Oh right! My bad. Will change that.

>     * I didn't use zimscraperlib because loading the .mo file from package data using `pkg_resources`, so its not tied to the directory structure. This should allow `warc2zim` to be run from any directory after being installed via pip -- I think zimscraperlib expects the `locale` directory to exist in current working directory.

Nope, you pass it the locale's parent folder and that's it. We're passing it the module folder so it would work anywhere it is installed.

ikreymer:
> 
> Nope, you pass it the locale's parent folder and that's it. We're passing it the module folder so it would work anywhere it is installed.

Ok, i was thinking to make it work with `zip_safe=True, but maybe that's not needed anymore when using pip.
It won't currently work with `python setup.py install` because there's no path, its loading from a .egg but maybe that behavior is obsolete now..

Also, I don't think it should thrown an error when the locale doesn't exist, i think it should just print warning and default to english (with null translations).. eg. if the default HTML pages is some other language

rgaudin:
> Ok, i was thinking to make it work with `zip_safe=True`, but maybe that's not needed anymore when using pip.
> It won't currently work with `python setup.py install` because there's no path, its loading from a .egg but maybe that behavior is obsolete now..

Yes, that's what I meant. made it zip_safe False.

> Also, I don't think it should thrown an error when the locale doesn't exist, i think it should just print warning and default to english (with null translations).. eg. if the default HTML pages is some other language

Agrees, forgot that behavior. Fixed it.

Issue #45, opened 2020-08-07
Errors loading on android


rgaudin:
Thanks to https://github.com/kiwix/kiwix-android/pull/2259, the SW is now running inside kiwix-android. Note that the main change is that the custom scheme and netloc is now `https://kiwix.app/`.
So a regular `index.html` article would be at `https://kiwix.app/A/index.html`

The basicsw.zim file loads properly. warc2zim created ones gets stuck at the loading message.

Here's what I get with regular code base (took the sw.js from latest PR though)

```
index.html:6 Inside index.html
index.html:8 window.mainUrl https://isago.ml/
VM17 load.js:1 inside load.js
VM17 load.js:3 inside load.js main()
VM17 load.js:9 SW support present
VM17 load.js:11 creating worker...
VM17 load.js:13 created Worker {onmessage: null, onerror: null}
VM17 load.js:15 window.location.href https://kiwix.app/A/index.html
VM17 load.js:19 Name kiwix.app
VM17 load.js:21 registering SW
sw.js:17 ww init
sw.js?replayPrefix=&root=kiwix.app:17 sw init
VM17 load.js:23 registered
VM17 load.js:33 posting message from worker
sw.js?replayPrefix=&root=kiwix.app:33 Activate!
VM17 load.js:26 worker received mesage MessageEvent {isTrusted: true, data: {…}, origin: "", lastEventId: "", source: null, …}
VM17 load.js:28 received collAdded going from https://kiwix.app/A/index.html to https://isago.ml/
(index):6 Inside index.html
(index):8 window.mainUrl https://isago.ml/
Navigated to https://kiwix.app/A/index.html
Navigated to https://kiwix.app/
```

Note the last two lines (not traces but browser's message). The browser gets directed to `https://kiwix.app/` instead of the correct URL.

At this point the SW looks properly registered for `kiwix.app/A`.

If I **manually** go to `https://kiwix.app/A/isago.ml/` (the homepage being `https://isago.ml/`), I get the homepage and everything works as expected.

![screenshot-2020-08-07_14 01 57 762](https://user-images.githubusercontent.com/57929/89643667-92fe2c80-d8b6-11ea-8d52-799faaa6e5bb.png)

I eventually figured out that the problem is with the `mainUrl` that isn't ready for such a URL and doesn't get the regular treatment it should. The following fixed it:

```javascript
window.location.replace("/A/" + window.mainUrl.replace("https://", ""));
```

[...]

Issue #43, opened 2020-08-05
Handle missing Service Worker scenarios


rgaudin:
With issues #10, #41 and in a way #24, we have important use cases broken because of the unavailability of the SW.

It seems logical a global approach would be used to solve them so here's the place to discuss what it could look like at both short and medium term.

A lot of discussion is happening on kiwixlib regarding the taskbar, raw content and APIs. It's important we track our options here with and without those features so we can participate in those discussions as well.

rgaudin:
## Contribution: inserted payload

That's the inelegant solution that first comes to mind.
The idea is to inject a `<script />` into every HTML article that will be executed on load.

This script would have to figure out whether it is being invoked from a SW-enabled environment or not.
If it is not, then it would redirect to a special *redirect page* that would register the SW and then display
the appropriate article with it.

I've been able to get a POC working so there should not be much trouble with that solution.

```javascript
async function main() {
    if (window.frameElement) {
        // we're in an iframe
        console.log("zim:: in iframe, doing nothing.")
        return;
    }

    // we're not, probably not registered
    console.log("zim:: not in iframe, redirecting...");
    window.location.href = "/redirect?" + window.location.href;
}();
```

**Pros**:
- Easy to implement

**Cons**:
- not elegant (editing each HTML article to add a script tag)
- requires `text/html` MIME on articles, so a libkiwix fix for the taskbar. kiwix/kiwix-lib#391 kiwix/kiwix-lib#394

rgaudin:
## Contribution: inserted payload via JS API

The (to-be-built) [JS API](https://wiki.openzim.org/wiki/Javascript_API) wants ZIM creators to be able to add a javascript file at a known location that would provide content-aware features to readers.
Those are not yet fully defined but we can imagine an *onload* endoint that would be executed each time an HTML article is loaded.

We could have the JS payload mentioned in the previous comment executed at this time.

_Note_: In the kiwix-serve context, the reader is kiwix-serve and would inject the JS API.

**Pros**:
- doesn't require us to edit content (readers will do that for us)

**Cons**:
- require JS API which is not defined and thus not implemented
- requires the `text/html` tweak also

ikreymer:
Since libkiwix already does injection into HTML, and since this needs to be changed to support raw output, I think best option would be to support this at the libkiwix level simply by supporting a different insert than the taskbar or no insert at all.

Here's one way that this could be done, without requiring a specific service worker implementation or iframes in libkiwix

[...]

Issue #41, opened 2020-08-05
Cold direct links are not working


rgaudin:
Accessing a direct link to a warc2zim-made article without going through the ZIM homepage first doesn't work : the browser offers to download the raw content as it is served as `application/octet-stream`.

The mime-type is purposely not `text/html` in order to prevent kiwix-serve from injecting the taskbar.

Although this visual consequence of appearing as a download can be fixed by changing the MIME-type ; it would not resolve this ticket as there is another problem:

The service worker is not registered and thus, if we had the correct MIME, the browser would only parse it and try to access any other resource (images, css, js, etc) from the actual remote URLs, resulting in either bandwidth usage or a broken page.

This is because the current code registers the SW in the homepage so skipping the homepage means no SW handling at all.

This cold-direct-link use case is **very important** and we can't change it because there are many solution integrator depending on it. The first ones coming to my mind are [kiwix-hotspot](https://github.com/kiwix/kiwix-hotspot), [Internet In a Box](https://github.com/iiab/iiab) and [OLIP](https://gitlab.com/bibliosansfrontieres/olip)

kelson42:
This is indeed an important problem we have here. For the moment, I don't believe there is an elegant way to fix the problem only by tweaking the ZIM content. Therefore I believe we well have somehow to define a formalism so the reader can know that there is something to absolutly do before rendering an HTML content (here secure that the SW is properly setup).

Issue #39, opened 2020-08-04
Use a visual loading feedback


rgaudin:
Following-up on #26, we should use a visual loader so we don't have to worry about translation regarding this step that every user will go through.

Any throbber would do, as long as it's discrete/light colors. @kelson42 suggested a spinner.

PR #47, opened 2020-08-11
Service Worker Fallback Handling

ikreymer:
Adds an injected 'catch all' handler that redirects to the index page if no service worker is installed (#43)

The injected template is in `sw_check.html`.

- If no service worker installed, redirect to index.html with `#redirect=` hash to indicate where to redirect to after installing the SW.
- The index page handles the error message already if service workers are not available.

Note: this is also assumes the raw `text/html;raw=true` support from kiwix/kiwix-lib#403.

The PR addresses:
- #45 - Better redirect from index page, detect prefix by /A/, should fix issue on Android
- #43 - Main SW issue
- #41 - Handles the cold direct links via redirect
- #10 - Handles force refresh
- #23 - 'Dead links' fall back to the kiwix-serve behavior of redirecting to live web or warning, with internal links showing a localized error message (could also be modified).
- #24 - Add additional warning message when loaded over HTTP
- #39 - Add a spinner instead of loading message.


src/warc2zim/templates/load.js
```diff
@@ -6,15 +6,27 @@ async function main() {
 
   var worker = new Worker("./sw.js");
 
-  const parts = window.location.href.split("/");
-  const inx = parts.indexOf("A");
-  const name = parts[inx - 1];
+  var prefix = window.location.href.slice(0, window.location.href.indexOf("/A/"));
```
rgaudin, line 9: That won't work on kiwix-serve if it's being used with a `--urlRootLocation` that contains a `/A/`.
Opening a separate issue for this as this might not be trivial to fix.

src/warc2zim/templates/sw_check.html
```diff
@@ -0,0 +1,22 @@
+<script>
+if (window.top === window && !window._WBWombat) {
+  if (!navigator.serviceWorker || !navigator.serviceWorker.controller) {
+    var inx =  window.location.href.indexOf("/A/");
```
rgaudin, line 4: idem

src/warc2zim/templates/load.js
ikreymer: Added better detection, now using regex to find a `/A/` before a string with `.`, so `http://example.com/A/path/A/example.com` will find `/A/example.com/` not `/A/path/A/example.com`

Of course, can't be entirely full proof.. if someone has `http://example.com/A/example.com/A/example.com/`
then this is ambiguous and not much we can do w/o specifying the prefix explicitly, and I hope that doesn't happen!

ikreymer:
Moving comment here so can see more easily:

Added better detection of prefix, now using regex to find a /A/ before a string with ., so http://example.com/A/path/A/example.com will find /A/example.com/ not /A/path/A/example.com

Of course, can't be entirely full proof.. if someone has http://example.com/A/example.com/A/example.com/
then this is ambiguous and not much we can do w/o specifying the prefix explicitly, and I hope that doesn't happen!

Also added a spinner for #39

Issue #49, opened 2020-08-13
Use zstd compression


rgaudin:
Use `compression="zstd"` on ZIM Creator so that warc2zim files are compressed with this algo.

Issue #14, opened 2020-07-27
Add codefactor


rgaudin:
* Add codefactor badge to README
* Fix [codefactor issues](https://www.codefactor.io/repository/github/openzim/warc2zim/issues)
* Make sure PR are blocked on codefactor issues

PR #50, opened 2020-08-19
flag improvements and docs:

ikreymer:
- name flags: use --name, --zim-file, --output convention for output, per spec in #4
- use zstd compression (#49)
- add codefactor badge to README (#14)
- update README with more info on domain filtering
- tests: use tmp_path fixture for temp dir


rgaudin:
Looks good (except the failing tests of course)

ikreymer:
Fixed :)

rgaudin (approved):

PR #54, opened 2020-08-19
release prep for 1.0.0:

ikreymer:
bump sw.js to latest 2.1.0-beta.0
bump version to 1.0.0
add github action to publish release on tag for #8
remove sw.js, update gitignore


.github/workflows/release.yaml
```diff
@@ -0,0 +1,34 @@
...
+        run: python -m pip install --upgrade pip wheel
+
+      - name: Build Dist
+        run: python setup.py sdist bdist_wheel
+
+      - name: Publish package
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_password }}
```
rgaudin, line 34: `${{ secrets.PYPI_API_TOKEN }}`

rgaudin: Can we use test.pypi for now (with `PYPI_TEST_API_TOKEN`) and once verified OK there change to regular pypi?

rgaudin (changes requested):
Thanks for the changes

Please:
* keep the whole thing on tag-only.
* Single publish block (I will comment test and uncomment correct token once OK)
* Change version to `0.0.2`.

I'll push a `0.0.2` tag and if that's all good on test.pypi, I'll fix it for regular pypi and push `0.1.0`.

Issue #59, opened 2020-09-25
Support validating arguments only (was: Create new --check-args mode)


ikreymer:
To support passing arbitrary arguments to warc2zim from zimit container (openzim/zimit#13), it might make sense to have
a 'check arguments' mode that can verify the arguments w/o doing anything just yet.

In this mode, `warc2zim --check-args <all other args>`, warc2zim will check for two things:
1) that all the arguments are valid
2) that the output directory is writable.

(This is almost like --dry-run but not quite, as it needs to check the directory).

This would allow running something like `docker run openzim/zimit --warc2zim "<all warc2zim args>"` which can then be passed to a warc2zim call on start-up.

This would prevent the crawl from starting if the arguments are invalid right away, or if output directory is not readable.
Otherwise, the crawl will run only to fail in the end due to invalid arguments, which is less desirable.

What do you think of this approach? @rgaudin

kelson42:
@ikreymer Why not just doing that per default at each run? Checking for input sanity seems anyway needed all the time.

ikreymer:
That's a good point, maybe it should not require WARCC inputs at all, and should succeed if params are valid and directory is writable?

Some examples would be:
`warc2zim` -- fail because `--name` is required
`warc2zim --name test` -- succeeds, if `/output` directory is writable, fail otherwise
`warc2zim --name test --output ./myoutput` -- succeeds if `./myoutput` exists and is writable

Or, should there still be a separate flag to check if output directory is writable?

kelson42:
@ikreymer Your last three examples seem to make sense to me. I like to print the usage() (like for --help) if not argument at all is given... but yes it should fail in that case. Everything in the input arguments which can easily be tested and avoid late exception should alwasy be tested at the very beginning, so we can later assume they are OK.

rgaudin:
OK, couple things I can think about:

- testing early so we can fail early is required.
- there aren't that much mandatory arguments (`--name` and `--output` for warc2zim and `--url` for zimit I believe).
- I don't like the `--warc2zim` approach as it moves some internal decisions up to the user. warc2zim should be abstracted there.
- help/usage is also required. It can be somewhat flexible. zimit usage can say `[other warc2zim args]` and instruct that those are available using `docker run openzim/zimit warc2zim --help`

So

- `warc2zim` should fail early on missing `--name` (I believe it's already the case) and on non-writable output (that's new).
- `zimit` should have a new thing that checks for `--name`, `--url` and `--output` (writable)

[...]

PR #60, opened 2020-10-06
Validate Inputs + Default Include All URLs

ikreymer:
Addresses input validation as described in #59: check output file is valid, exit with error code 100 if everything is valid but no inputs

Also addresses openzim/zimit#12 -- default to including all URLs unless using `--include-domains` flag,
simplifies configuration.

Update version to 1.1.0 due to breaking changes, removing `-a`/`--include-all` flag, which is now the default.


rgaudin (commented):
Couple of issues:

- returncode is not set on main.py so it doesn't work when testing locally.
- there's no changelog entry
- you're doing exactly what you're trying to avoid: creating a file when you don't want to (when `touch`ing the output filename)

I've fixed all those so we don't waste a round-trip review on this. If that's not OK with you, we'll revert.

PR #62, opened 2020-10-19
improvements to better support zimit: (openzim/zimit#29)

ikreymer:
- add support for directory inputs, processing all .warc and .warc.gz files recursively in a directory
- fix trailing slash: if `https://example.com/` is specified, treat as `http://example.com/`
- bump version to 1.2.0


rgaudin (approved):
Approving this so we can release but we'll need to raise the test coverage

PR #64, opened 2020-10-21
Support extra fuzzy matching/canonicalization of certain URLs

ikreymer:
To support video replay openzim/zimit#9 (in particular, youtube), it is necessary to create extra entries for 'inexact matching' of URLs to those with known parameters. (wabac.js / replayweb.page also handles this via prefix search, but since zim does not support prefix search currently, creating extra redirect entries is needed).

The approach is to check if a URL has an inexact/fuzzy matching rule, and if so, add a redirect from that rule to the exact URL. The redirect is done via a redirect record in `H/<fuzzy url>` to match other revisit records.
(ZIM redirects are not used since the service worker still needs to be able to load the `H/` record and not a redirect)

Tests and a sample ZIM are added


rgaudin:
I am not sure I understand properly what is going on.

Looking at the code, it seems that you are creating a revisit record for url `youtube.fuzzy.replayweb.page` (`fuzzy_url`) that *points* (refers-to) `xxxxxgooglevideos` (`url`). I would expect the opposite: crawled URL (`url`) to redirect to `fuzzy` one that the replay system can catch and interpret differently.

How is the fuzzy URL ever hit in your scenario? Is there some kind of matching between webac.js that would do this fuzzy matching in JS and then query the fuzzy'd url in the ZIM?
If so, this needs to be properly documented in the code otherwise this part will soon become a blind unmaintained hazard :)

ikreymer:
Yes, there should be a more detailed explanation somewhere perhaps, currently there are replay rules in wabac.js, as well as in pywb.

The basic idea is that the URL requested at capture time and crawl time is different, eg. a video might be captured from `https://some-domain.googlevideo.com/videoplayback?someparams=abc&id=1234&someotherparams=def`
But when replaying it, it may request:
`https://some-other-domain.googlevideo.com/videoplayback?someparams=def&id=1234&someparams=ghi`

If there is no exact match, both URLs are canonicalized to: `https://youtube.fuzzy.replayweb.page/videoplayback?id=1234`

For example:
1) request is to: `https://some-other-domain.googlevideo.com/videoplayback?someparams=def&id=1234&someparams=ghi`, no match.
2) look for: `https://youtube.fuzzy.replayweb.page/videoplayback?id=1234`, find revisit record.
3) the revisit redirects to: `https://youtube.fuzzy.replayweb.page/videoplayback?id=1234` -> `https://some-domain.googlevideo.com/videoplayback?someparams=abc&id=1234&someotherparams=def` which is the URL
actually captured.

This is why the reference is needed from the fuzzy url to the original.

This is a bit limited by ZIM not supporting prefix queries, otherwise another, another more flexible option is to do a prefix search, as that allows for partial matching in cases where there are multiple parameters. Another common example is something like: captured `https://example.com/?a=b&c=d&_=123456` vs `https://example.com/?a=b&c=d&_=123457` where the _= param can be ignored, or a best match can be found automatically, but that requires being able to search by prefix, eg. for all URLs starting with `https://example.com/?`

rgaudin (approved):

PR #76, opened 2020-12-30
Better missing service worker error page

### LINKED ISSUE #73: Change the sentence to have a link pointing to localhost (local kiwix-serve use case).
This page must be loaded via an HTTPS URL to support service workers.
Try Loading HTTPS URL?

kelson42:
Fixes #73


rgaudin (changes requested):
Thank you for your PR.
Unfortunately, this is not merge-able:

- you are trying to insert a bunch of elements inside an `<h3 />` one. See `templates/index.html`.
- You are inserting those as text instead of actual elements
- you did not update the catalog template
- you did not submit the compiled version of the translation

If you eventually change the `<h3 id="error" />` for another element, please change the behavior of `404.html`  as well as it expects an header I think.

rgaudin (approved):

PR #78, opened 2021-01-12
Add support for --custom-css

### LINKED ISSUE #61: Tweak the rendering with custom CSS
Now that we have a few ZIM files made via ZIMit/warc2zim, we can see how well the scraping and ZIM file looks like. To some extend this is too accuratly mimicing the original web site even with things which won't work and will never work like formular (leave comment, search, etc...) and even a few other things which might be even simply not welcome.

I propose to allow to put a custom CSS which would apply to each page in the ZIM. We could do that with a parameter called `--customCss=[path|url]`.

For the moment, not a priority, just asking to (in)valide the idea.

rgaudin:
- new --custom-css option taking a path or a URL
- converted to a WARC record on a fixed URL: https://warc2zim.kiwix.app/custom.css
- stylesheet link added at very end of head for every HTML article

Fixes #61 

Reviewers: look at second commit only; first one is code-formatting

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,10 @@ To simply include all urls, use the `--include-all` / `-a` flag:
 warc2zim myarchive.warc --name myarchive -a -u https://someother.example.com/page.html
 ```
 
+### Custom CSS
+
+`--custom-css` allows passing an URL or a path to a CSS file that gets added to the ZIM and gets included on **every HTML article** at the very end of `</head>` (if it exists).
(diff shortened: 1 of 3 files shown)


rgaudin:
@kelson42 ready to review now.

kelson42 (approved):
@rgaudin Great, important that we get this in the Zimfarm as well, so I can start to play with it.

PR #127, opened 2023-11-16
Fix url filtering doc

### LINKED ISSUE #126: Documentation in README.md about URL filtering is wrong
Documentation in [README.md about URL filtering](https://github.com/openzim/warc2zim/#url-filtering) is wrong because it does not reflect the fact that since 1.1.0 and [commit 48bb2ed](https://github.com/openzim/warc2zim/commit/48bb2ed9a0b71f92b398f7347b34167533a14b7c) the default behavior is to include all URLs unless --include-domains flag is specified.

benoit74:
Fix #126 

Changes:
- fix documentation about URL filtering
- fix wrong titles levels

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ warc2zim ./path/to/myarchive.warc --output /output --name myarchive.zim -u https
 
 The above will create a ZIM file `/output/myarchive.zim` with `https://example.com/` set as the main page.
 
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
LGTM ; I don't think I've ever used that option 😅

PR #129, opened 2023-11-27
Split the one big source code file into small submodules.

mgautierfr:


mgautierfr:
It is just a reorganization (preparatory work for next PRs). No functional changes.

rgaudin:
There are many things that are unappealing in this but it's mostly (if not completely) from previous version.

I would not want to slow down the zimit2 rewrite with changes that are mostly setup/maintenance related. @benoit74 and I now have a quite standard way of doing python projects but it's an extensive effort to adapt to it for a third party.

If you agree, I'd suggest we merge this refactoring and review zimit2-related PRs for code (and not setup/convention) and then only, @benoit74 and I can tweak it so it's in line with other projects (that's already overdue).

@mgautierfr can you just make the following changes:

- use absolute imports (`from warc2zim.convert import Converter` instead of `from .converter import Converter`)
- use explicit imports. In main, you import `get_version` from `converter` while it is declared in `utils`.

benoit74:
Thank you @rgaudin for clarifying my embarrassment with this PR. I'm fine with your suggestion.

@mgautierfr: of course we do have documentation 🤣 : https://github.com/openzim/_python-bootstrap/wiki/

rgaudin:
> > @benoit74 and I now have a quite standard way of doing python projects but it's an extensive effort to adapt to it for a third party.
> 
> Do you have a documentation of the "quite standard way" ? (Genuine question)

https://github.com/openzim/_python-bootstrap serves as a bootstrap/boiler plate for python projects. We've settled a lot of conventions into it so complying projects are easier to maintain

mgautierfr:
A small up 😉 
Are you ok with the PR ?

benoit74 (approved):

PR #130, opened 2023-12-08
Introduce normalize and normalization schema.

mgautierfr:
This is a first PR with "real" work on warc2zim2.

- Introduce url normalization (in place of "canonicalize")
- Don't store headers
- Remove all service worker stuff (`sw.js` itself, `index.html`, ...)

This PR should works with website which don't need any content or url rewriting. Meaning:
- Local resource only.
- All links relative
- No js or "simple" js (following the above rules)
- No url being reduced (or else, with link pointed already to the reduced url)


src/warc2zim/converter.py
```diff
@@ -64,11 +57,8 @@
 # HTML mime types
 HTML_TYPES = ("text/html", "application/xhtml", "application/xhtml+xml")
 
-# external sw.js filename
-SW_JS = "sw.js"
-
 # head insert template
-HEAD_INSERT_FILE = "sw_check.html"
+HEAD_INSERT_FILE = None
```
benoit74, line 61: Any reason to keep all this stuff about head inserts? It looks they are not used anymore, but do you keep them because you will need them at a later stage? If we are mostly sure they won't be needed, I would get rid of them (we can always come back to this commit to re-add them if needed). I generally dislike dead code which often deprecates very fast and gives a false sense of "we just have to reactivate the feature".

benoit74 (changes requested):
Mainly one "stupid" question. 
I cannot really review the rest, it is a too hard for me to follow where we are going. At least I don't see things which looks weird to me.

mgautierfr:
We now use the `addAlias` method, which need the (really) recent version of libzim (PR https://github.com/openzim/python-libzim/pull/175 not yet merged)

And we need to update https://github.com/openzim/python-scraperlib/ as it is through this project that we get `libzim` deps.

benoit74:
As we said last monday, we should probably release all this stuff so that the CI and us can test this correctly.

And again, we should not merge this to `main` but to a temporary branch.

kelson42:
@mgautierfr Please create a `zimit2` branch to gather all your work.

rgaudin (changes requested):
Thank you ; given we're very close to a libzim release, I didn't go through the trouble of building a special pylibzim to run the tests.

Only a few minor comments on my side 👍

mgautierfr:
Two "fixup" commits added in the git history.

rgaudin (approved):
All good now 👍

kelson42:
Can we rely on new’y released pythin-libzim 3.4.0 and merge?

rgaudin:
No it depends on scraperlib

kelson42:
@rgaudin Then https://github.com/openzim/python-scraperlib/issues/116 is the blocker

rgaudin:
`3.2.0` released

kelson42:
The CI still fails! Probably nothing really bad, but we have to fix in priority to move forward.

Jaifroid:
Just to double-check: has it really been decided that Headers should never be stored, given that they provide instructions to the browser about how to deal with Responses (see https://github.com/kiwix/overview/issues/95#issuecomment-1828176533)? In current warc2zim-generated archives, every Response is constructed from two lookups: a first-pass lookup to get the Response Header, and a second-pass lookup to get the Response Body. Only when both lookups have been done, does the backend have enough information to send a complete Response (mirroring that which was originally recorded) back to the browser. I fear (I hope the fear is unfounded, but I have some doubts) that if the functionality is removed, we're going to have trouble with replaying highly dynamic sites (as you know, it's pretty fragile as it is).

kelson42:
> Just to double-check: has it really been decided that Headers should never be stored

We will implement proper headers storage and mgmt only if necessary to provide nice replay feature. So far, it seems this is not necessary. But, it might be implemented in the future.

mgautierfr:
Rebased fixup. Ready to merge.

PR #135, opened 2023-12-18
Fix entry_points.

mgautierfr:
In commit 15567179 we have changed the function name to `main`.


rgaudin (approved):
LGTM ; this will require a change on zimit I think

PR #139, opened 2023-12-19
Add js rewriting and wombat.

### LINKED ISSUE #136: Rewrite js content for warc2zim without SW

mgautierfr:
Next step in our warc2zim2 branch : Rewrite js content (statically and dynamically)

Now, we should be able to have dynamic websites (including "complex" js to work).

Fix #136

### DIFF
--- a/src/warc2zim/content_rewriting/__init__.py
+++ b/src/warc2zim/content_rewriting/__init__.py
@@ -1,2 +1,3 @@
 from warc2zim.content_rewriting.html import HtmlRewriter
 from warc2zim.content_rewriting.css import CSSRewriter
+from warc2zim.content_rewriting.js import JSRewriter
--- a/src/warc2zim/content_rewriting/__init__.py
+++ b/src/warc2zim/content_rewriting/__init__.py
@@ -1,3 +1,3 @@
 from warc2zim.content_rewriting.html import HtmlRewriter
-from warc2zim.content_rewriting.css import CSSRewriter
-from warc2zim.content_rewriting.js import JSRewriter
+from warc2zim.content_rewriting.css import CssRewriter
+from warc2zim.content_rewriting.js import JsRewriter
--- a/src/warc2zim/content_rewriting/__init__.py
+++ b/src/warc2zim/content_rewriting/__init__.py
@@ -1,0 +1,2 @@
+from warc2zim.content_rewriting.html import HtmlRewriter
+from warc2zim.content_rewriting.css import CSSRewriter
--- a/src/warc2zim/content_rewriting/js.py
+++ b/src/warc2zim/content_rewriting/js.py
@@ -219,6 +219,9 @@ def create_js_rules() -> List[TransformationRule]:
     ]
 
 
+REWRITE_JS_RULES = create_js_rules()
+
+
 class JsRewriter:
     """
     JsRewriter is in charge of rewriting the js code stored in our zim file.
@@ -289,7 +292,7 @@ class JsRewriter:
         if not opts.get("isModule"):
             opts["isModule"] = self._detect_is_module(text)
 
-        rules = create_js_rules()
+        rules = REWRITE_JS_RULES[:]
 
         if opts["isModule"]:
             rules.append(self._get_esm_import_rule())
--- a/src/warc2zim/content_rewriting/js.py
+++ b/src/warc2zim/content_rewriting/js.py
@@ -3,20 +3,25 @@ from typing import Any, Callable, Optional, Iterable, Tuple, Mapping, List
 
 TransformationRule = Tuple[re.Pattern, Callable[[re.Match], str]]
 
-
+# Regex used to check if we ar import or exporting things in the string
+# ie : If we are a module
 IMPORT_RX = re.compile(r"""^\s*?import\s*?[{"'*]""")
 EXPORT_RX = re.compile(
     r"^\s*?export\s*?({([\s\w,$\n]+?)}[\s;]*|default|class)\s+", re.M
 )
 
+# The regex used to rewrite `import ...` in module code.
 IMPORT_MATCH_RX = re.compile(
     r"""^\s*?import(?:['"\s]*(?:[\w*${}\s,]+from\s*)?['"\s]?['"\s])(?:.*?)['"\s]""",
 )
 
+# A sub regex used inside `import ...` rewrite to rewrite http url imported
 IMPORT_HTTP_RX = re.compile(
     r"""(import(?:['"\s]*(?:[\w*${}\s,]+from\s*)?['"\s]?['"\s]))((?:https?|[./]).*?)(['"\s])""",
 )
 
+# This list of global variables we want to wrap.
+# We will setup the wrap only if the js script use them.
 GLOBAL_OVERRIDES = [
     "window",
     "globalThis",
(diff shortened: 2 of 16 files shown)


src/warc2zim/content_rewriting/__init__.py
```diff
@@ -0,0 +1,2 @@
+from warc2zim.content_rewriting.html import HtmlRewriter
+from warc2zim.content_rewriting.css import CSSRewriter
```
benoit74, line 2: For style and consistency with our other rewriters, `CssRewriter` (`ss` in lower case) is probably more appropriate than `CSSRewriter`

src/warc2zim/templates/head_insert.html
```diff
@@ -0,0 +1,207 @@
...
+  // Use to timestamp message send to top frame. Don't care
+  wbinfo.timestamp = "";
+
+  // Use to send message to top frame and in default rewrite url fonction. Don't care
+  wbinfo.request_ts = "";
+
+  // The url only which we are served.
+  wbinfo.prefix = current_url.substring(
+    0,
+    current_url.length - "{{ path }}".length,
```
rgaudin, line 152:
```suggestion
    current_url.length - ("{{ path }}".length),
```
better be safe with JS!

src/warc2zim/templates/head_insert.html
```diff
@@ -0,0 +1,207 @@
...
+
+<!-- WB Insert -->
+<script>
+  // Get the current url we are serving from.
+  // We cannot know it at zim creation as it depends of our server.
+  const current_url = (function () {
+    var href = new URL(window.location.href);
+    href.hash = "";
+
+    return String(href);
```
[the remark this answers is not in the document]
rgaudin: and elsewhere

src/warc2zim/items.py
```diff
@@ -8,50 +8,58 @@
...
             self.content = self.record.content_stream().read()
 
+        should_rewrite = getattr(record, "method", "GET") != "POST"
+        if not should_rewrite:
+            return
+
         if self.mimetype.startswith("text/html"):
+            wombat_path = url_rewriter.from_normalized("_zim_static/wombat.js")
+            head_insert = head_template.render(
+                path=path, wombat_path=wombat_path, url=get_record_url(record)
```
rgaudin, line 54: Ain't it missing `orig_scheme` and `orig_host` (and maybe others)?

tests/test_js_rewriting.py
```diff
@@ -0,0 +1,226 @@
+import pytest
+from warc2zim.content_rewriting import JSRewriter
+from warc2zim.url_rewriting import ArticleUrlRewriter
+from textwrap import dedent
+
+
+def wrapScript(text: str) -> str:
```
rgaudin, line 7: dosctring

src/warc2zim/items.py
```diff
@@ -44,6 +44,10 @@ def __init__(
         else:
             self.content = self.record.content_stream().read()
 
+        should_rewrite = getattr(record, "method", "GET") != "POST"
```
benoit74, line 47: Should we not rewrite POST requests or only rewrite GET? I don't know how many PATCH/PUT/DELETE requests we might have (probably rare) but I don't see why they should be more rewriten than POST requests

benoit74 (commented):
Wombat is definitely not something I'm an expert of, so only minor comments from my side.

src/warc2zim/items.py
mgautierfr: Yes indeed. I missed a commit during the cleanup of the git history.
See the two last commits.

mgautierfr:
I have take into account all your comments, but the two last commits are important as they are code I've missed during the cleanup of the git history before create PR.

benoit74:
Nothing to add on the last two commits (but I agree they are probably important ^^)

mgautierfr:
Testing of zim files generated by this PR will need this [patch](https://github.com/kiwix/libkiwix/pull/1036) for libkiwix/kiwix-serve

src/warc2zim/items.py
```diff
@@ -67,13 +75,13 @@ def __init__(self, env, filename, main_url, **kwargs):
         self.mime = get_mime_for_name(filename)
         self.mime = self.mime or "application/octet-stream"
 
-        if filename != SW_JS:
-            template = env.get_template(filename)
-            self.content = template.render(MAIN_URL=self.main_url)
-        else:
+        if kwargs.get("raw", False):
```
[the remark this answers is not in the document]
mgautierfr: Resolved by removing add template as we don't have any templates to directly store 
as entry.

tests/test_js_rewriting.py
mgautierfr: Fixed

src/warc2zim/items.py
mgautierfr: I agree. In fact, I think the condition is not right at all.
We should not rewrite content designated to be use by js (as ajax call response will already be rewritten by wombat).

However it seems their is not standard way to detect a request come from ajax. Maybe we have to work with the scrapper side (you) to mark records as coming from ajax (or simply as "don't rewrite").

But I propose to move this in another issue as we will probably don't fix this easily.

rgaudin: We had this exact discussion on our call with Ilya. I remember there's no way to know it from the WARC but we envisioned a similar solution: have browsertrix add this information to it during crawl. The information is in the browser (even in the dev tools) but I don't know if it's exposed in their API and how easy it would be to implement.
Once you've opened the ticket here, we'll open a ticket upstream (browsertrix-crawler).

mgautierfr: Issue created : https://github.com/openzim/warc2zim/issues/140

src/warc2zim/templates/head_insert.html
```diff
@@ -0,0 +1,207 @@
+{% autoescape false %} {% set urlsplit = url | urlsplit %}
```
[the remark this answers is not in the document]
mgautierfr: I agree. I have a better version but I think it would be better in another PR.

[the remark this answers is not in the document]
mgautierfr: New PR https://github.com/openzim/warc2zim/pull/141 creating for this rewrite.

src/warc2zim/content_rewriting/js.py
```diff
@@ -0,0 +1,252 @@
...
+    "self",
+    "document",
+    "location",
+    "top",
+    "parent",
+    "frames",
+    "opener",
+]
+
+GLOBALS_CONCAT_STR = "|".join(
```
rgaudin, line 32: Not sure it has value on its own. Given it's already hardly readable, I think it can be combined with `GLOBALS_RX` directly

rgaudin (changes requested):
Thank you ; it looks good expect what I listed in my previous comment:

- wombat should be treated as the external dependency it is.
- js.py should be improved for clarity and maintenance: dosctrings, snake case, comments, variable names, no all-caps vars inside method

mgautierfr:
Last commit add a lot of comments in `content_rewriting/js.py`.

> wombat should be treated as the external dependency it is.

There is currently no `wombat.js` (with needed patch) published somewhere.
I've open the PR https://github.com/webrecorder/wombat/pull/128 to apply the patch upstream.
When accepted, and wombat released, we could use the same mechanism as for wabac.js to download wombat.
So I propose that we move this into a new issue.

rgaudin:
> There is currently no `wombat.js` (with needed patch) published somewhere. I've open the PR [webrecorder/wombat#128](https://github.com/webrecorder/wombat/pull/128) to apply the patch upstream. When accepted, and wombat released, we could use the same mechanism as for wabac.js to download wombat. So I propose that we move this into a new issue.

Ah OK I didn't realize if was patched

src/warc2zim/content_rewriting/js.py
```diff
@@ -29,79 +34,141 @@
...
+
         @m2str
         def f(x):
             return x.replace(src, target)
 
         return f
 
-    def replaceThisProp():
-        def f(m_object, opts):
+    def replace_this_prop():
```
rgaudin, line 146:
```suggestion
    def replace_this_non_prop():
```

src/warc2zim/content_rewriting/js.py
```diff
@@ -29,79 +34,141 @@
...
+        """
+        Create a rewrite_function replacing "this" by `this_rw`.
+
+        Replacement appears only if "this" is not a property of a object.
```
rgaudin, line 150:
```suggestion
        Replacement happen only if "this" is not a property of an object.
```

src/warc2zim/content_rewriting/js.py
```diff
@@ -111,128 +178,147 @@ def f(m_object, opts):
...
         ),
     ]
 
 
 class JsRewriter:
+    """
+    JsRewriter is in charge of rewriting the js code stored in our zim file.
+
+    The main "input" is a list of rules, each rule being a tuple (regex, rewriting_function).
+    We want to applay each rule to the content. But doing it blindly is conter productives.
```
rgaudin, line 227:
```suggestion
    We want to apply each rule to the content. But doing it blindly is counter-productive.
```

src/warc2zim/content_rewriting/js.py
```diff
@@ -111,128 +178,147 @@ def f(m_object, opts):
...
 
-    def initLocalDeclaration(self, localDecls: Iterable[str]) -> str:
-        ASSIGN_FUNC = "_____WB$wombat$assign$function_____"
-        buffer = f"""var {ASSIGN_FUNC} = function(name) {{return (self._wb_wombat && self._wb_wombat.local_init && self._wb_wombat.local_init(name)) || self[name]; }};
-if (!self.__WB_pmw) {{ self.__WB_pmw = function(obj) {{ this.__WB_source = obj; return this; }} }}
-{{
-"""
-        for decl in localDecls:
-            buffer += f"""let {decl} = {ASSIGN_FUNC}("{decl}");\n"""
+    def _init_local_declaration(self, local_decls: Iterable[str]) -> str:
```
rgaudin, line 245: Maybe a docstring stating that we always start with this boilerplate code

rgaudin (approved):
OK thank you ; it's a lot more readable and understandable 👍

src/warc2zim/content_rewriting/js.py
mgautierfr: Indeed. Even more as we don't always start with this boilerplate code.

mgautierfr:
Already approved commits are rebased-fixup.
Only the two last (which are pretty straightforward) to review.

rgaudin:
LGTM

PR #141, opened 2023-12-20
Better/Simple(/Stronger) wombat setup.

mgautierfr:
No real functional change. Only code polishing.

Edit: Important code polishing, we now have removed the duplicated definition of fuzzy rules.

### DIFF
--- a/src/warc2zim/statics/wombat_setup.js
+++ b/src/warc2zim/statics/wombat_setup.js
@@ -3,89 +3,89 @@
 // and returns a configuration dictionnary to pass to wombat.
 
 const getWombatInfo = (function () {
-// Reduce "complex" url to our simplyfied ones.
-// ! MUST be in sync with the python code (at zim creation time)
-let reduce = function (path) {
-  const fuzzy_rules = [
-    {
-      match: new RegExp(
-        ".*googlevideo.com/(videoplayback\\?).*((?<=[?&])id=[^&]+).*",
-      ),
-      replace: "youtube.fuzzy.replayweb.page/$1$2",
-    },
-    {
-      match: new RegExp(
-        "(?:www\\.)?youtube(?:-nocookie)?\\.com/(get_video_info\\?).*(video_id=[^&]+).*",
-      ),
-      replace: "youtube.fuzzy.replayweb.page/$1$2",
-    },
-    {
-      match: new RegExp("(\\.[^?]+\\?)[\\d]+$"),
-      replace: "$1",
-    },
-    {
-      match: new RegExp(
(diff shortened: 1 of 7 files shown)


kelson42:
@mgautierfr Why the fuzzy rules, both the one for Python and the one in javascript, are not data-driven? That would allow:
* to avoid sync errors between the two world
* easy update
* easy/proper tracking of the rules

src/warc2zim/statics/wombat_setup.js
```diff
@@ -0,0 +1,197 @@
...
+    for (const rule of fuzzy_rules) {
+      const new_path = path.replace(rule.match, rule.replace);
+      if (new_path != path) {
+        return new_path;
+      }
+    }
+    return path;
+  };
+
+  const getWombatInfo = function (
```
benoit74, line 63: Having a variable named `getWombatInfo` inside another variable named `getWombatInfo` does not really help in readability, it took me a while to understand this relation.

src/warc2zim/statics/wombat_setup.js
```diff
@@ -0,0 +1,197 @@
+// This file setup a wbinfo dictionnary to configure a wombat instance.
+// It mainly export one function `getWombatInfo` which takes context information
+// and returns a configuration dictionnary to pass to wombat.
+
+const getWombatInfo = (function () {
+  // Reduce "complex" url to our simplyfied ones.
+  // ! MUST be in sync with the python code (at zim creation time)
+  let reduce = function (path) {
```
benoit74, line 8: I don't get why this is not a `const`?

benoit74 (changes requested):
Some minor changes + I second the question of @kelson42 about why the wombat_setup.js is not a template solved by Jinja. so that we can declare all fuzzyrules in a more "data-driven" and unique place.

I did not added again all remarks made in previous PR by renaud ^^

src/warc2zim/statics/wombat_setup.js
mgautierfr: This is now replaced by a "direct" return:

```js
return function(
...
```

src/warc2zim/statics/wombat_setup.js
mgautierfr: Keyboard mistake. I swear (please trust me), I've typed `const` but the keyboard writes `let` instead.

This is fixed.

mgautierfr:
> Why the fuzzy rules, both the one for Python and the one in javascript, are not data-driven? 

This is now the case with the last commit.

benoit74 (approved):
LGTM, thank you

PR #142, opened 2023-12-21
Handles external links (without SW)

### LINKED ISSUE #137: Handle external links in warc2zim without SW

mgautierfr:
Fixes #137 

With this PR we don't rewrite html (`a href`) link when the url is not in the warc archive.

Other links are always rewritten to avoid zim file silently phoning home.

### DIFF
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -488,7 +488,11 @@ class Converter:
                 return
 
             payload_item = WARCPayloadItem(
-                normalized_url, record, self.head_template, self.css_insert
+                normalized_url,
+                record,
+                self.head_template,
+                self.css_insert,
+                self.existing_entries,
             )
 
             if len(payload_item.content) != 0:
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -132,6 +132,7 @@ class Converter:
 
         self.indexed_urls = set({})
         self.revisits = {}
+        self.existing_entries = set({})
 
         # progress file handling
         self.stats_filename = (
@@ -206,7 +207,7 @@ class Converter:
             )
             return 100
 
-        self.find_main_page_metadata()
+        self.gather_information_from_warc()
         self.title = self.title or "Untitled"
         if len(self.title) > 30:
             self.title = f"{self.title[0:29]}…"
@@ -297,16 +298,23 @@ class Converter:
 
         yield from iter_warc_records(self.inputs)
 
-    def find_main_page_metadata(self):
-        for record in self.iter_all_warc_records():
+    def gather_information_from_warc(self):
+        main_page_found = False
+        for record in iter_warc_records(self.inputs):
+            url = get_record_url(record)
+            normalized_url = normalize(url)
+
+            self.existing_entries.add(normalized_url)
+
+            if main_page_found:
+                continue
+
             if record.rec_type == "revisit":
                 continue
 
             # if no main_url, use first 'text/html' record as the main page by default
             # not guaranteed to always work
             mime = get_record_mime_type(record)
-
-            url = record.rec_headers["WARC-Target-URI"]
 
             if (
                 not self.main_url
@@ -317,9 +325,9 @@ class Converter:
                     or record.http_headers.get_statuscode() == "200"
                 )
             ):
-                self.main_url = normalize(url)
+                self.main_url = normalized_url
 
-            if urldefrag(self.main_url).url != normalize(url):
+            if urldefrag(self.main_url).url != normalized_url:
                 continue
 
             # if we get here, found record for the main page
@@ -331,7 +339,8 @@ class Converter:
                     "Main page is not an HTML Page, mime type is: {0} "
                     "- Skipping Favicon and Language detection".format(mime)
                 )
-                return
+                main_page_found = True
+                continue
 
             record.buffered_stream.seek(0)
             content = record.buffered_stream.read()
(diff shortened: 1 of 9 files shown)


src/warc2zim/converter.py
```diff
@@ -282,17 +283,24 @@ def iter_all_warc_records(self):
 
         yield from iter_warc_records(self.inputs)
 
-    def find_main_page_metadata(self):
-        for record in self.iter_all_warc_records():
+    def gather_information_from_warc(self):
```
rgaudin, line 286: I think the name is misleading. It's all about the main page (and its metadata) and given how long this is already, probably good not to add additional stuff to it

src/warc2zim/converter.py
```diff
@@ -130,6 +130,7 @@ def __init__(self, args):
 
         self.indexed_urls = set({})
         self.revisits = {}
+        self.existing_entries = set({})
```
rgaudin, line 133: I think it's more of a `warc_urls` than `existing_entries` which is unclear about whether processed or not and whether thoses are ZIM entries or WARC entries and given it's only URLs, entries itself is misleading

src/warc2zim/items.py
```diff
@@ -30,7 +31,12 @@ class WARCPayloadItem(StaticItem):
     """
 
     def __init__(
-        self, path: str, record: ArcWarcRecord, head_template: str, css_insert: str
+        self,
+        path: str,
+        record: ArcWarcRecord,
+        head_template: str,
+        css_insert: str,
+        known_urls: Set[str],
```
rgaudin, line 39: See ? Here' you're calling it `known_urls` 😉

src/warc2zim/url_rewriting.py
```diff
@@ -135,17 +135,23 @@ def normalize(url: str | bytes) -> str:
     return path
 
 
+def remove_anchor(url: str) -> str:
```
rgaudin, line 138:
```suggestion
def get_with_anchor(url: str) -> str:
```

Since you're returning a modified version and not modifying in-place. I like _fragment_ better than anchor but both are understandable

tests/test_url_rewriting.py
```diff
@@ -12,7 +12,7 @@
     ]
 )
 def rewriter(request):
-    yield ArticleUrlRewriter(request.param)
+    yield ArticleUrlRewriter(request.param, set(["kiwix.org/bar/foo"]))
```
rgaudin, line 15:
```suggestion
    yield ArticleUrlRewriter(request.param, {"kiwix.org/bar/foo"})
```

tests/test_html_rewriting.py
```diff
@@ -163,7 +172,9 @@ def test_extract_title():
 
 
 def test_rewrite_attributes():
-    rewriter = HtmlRewriter("kiwix.org/", "", "")
+    rewriter = HtmlRewriter(
+        ArticleUrlRewriter("kiwix.org/", set(["kiwix.org/foo"])), "", ""
```
rgaudin, line 176:
```suggestion
        ArticleUrlRewriter("kiwix.org/", {"kiwix.org/foo"}), "", ""
```

tests/test_html_rewriting.py
```diff
@@ -144,7 +145,15 @@ def rewrite_url(request):
 
 def test_rewrite(rewrite_url):
     assert (
-        HtmlRewriter(rewrite_url.article_url, "", "").rewrite(rewrite_url.input).content
+        HtmlRewriter(
+            ArticleUrlRewriter(
+                rewrite_url.article_url, set(["exemple.com/a/long/path"])
```
rgaudin, line 150:
```suggestion
                rewrite_url.article_url, {"exemple.com/a/long/path"}
```

rgaudin (changes requested):
Good!

rgaudin:
> I'm also concerned by the fact that we keep all know URLs in memory. 

This is not a new concern ; several scrapers need to store huge list of entries/urls. There's no average size for a URL but even in UTF-8, most characters in most URLs are ASCII so I think 500b for an average URL can be seen consevative. With a hundred thousands of them it's _just_ over 50MB. Sure that's a lot but in the context of a scraper turning 100K requests into a ZIM, that's not an issue IMO.

src/warc2zim/converter.py
mgautierfr: The problem is that doing the main page searching and the filling of known_url in two different steps means that we loop twice (and one more time for the real processing) the warc entries.

rgaudin: I am not suggesting we change the behavior ; just the name. Maybe `find_homepage_and_urls` or something

mgautierfr:
> I find it's very weird to do so many things in the WARCPayloadItem class __init__ method, especially rewriting item URLs in this class seems odd. This is made especially visible by the fact that we need to pass all known URLs to the class, which means this class knows a lot of things about the surrounding logic. I would have expected this class to be more a kind of adapter to some ZIM specifities, based on pre-processed information.

I partly agree. This `__init__` method become a bit to big.
We could move the rewrite part in a separated function, but then it would be this function to knows a lot of things about the surrounding logic. And either we pass the information to the `__init__` which will call the rewrite function or we could the rewrite function in the converter and pass the rewriten content to the item. In the first case, it doesn't help a lot. In the second case, we could simply drop `WARCPayloadItem` and simply use `StaticItem`.

> I'm also concerned by the fact that we keep all know URLs in memory. We've already seen websites with more than 10k pages, and I would expect we will have to deal with 100k pages at some point, meaning 100k URLs in memory. I don't know exactly how bad this could be (in terms of memory footprint), but I think we should be at least aware of this / track this risk/need in a specific issue.

Agree. But as has said @rgaudin, it should not be such a problem. And I don't know how to do it without storing the urls somewhere (or loop the whole warc for each url rewrite)

src/warc2zim/converter.py
mgautierfr: Renamed to `warc_urls`

rgaudin (approved):
LGTM

benoit74 (approved):
LGTM

I agree the problem I mentioned is not such of an issue. A simple solution (I didn't say straightforward) would have been to store all data in a local DB, typically SQLite like is done for Gutenberg. Of course it comes with some drawbacks it terms of CPU / scrape duration, but at least it is scalable.

And you convinced me regarding the `__init__` method, let's keep it like this until we find it becomes a real issue.

mgautierfr:
(Put in draft as it needs merging of #141 first)

PR #146, opened 2024-01-12
Download released wombat.js instead of embed it in our respository.

### LINKED ISSUE #143: Treat Wombat as a dependency
Wombat.js should not be included in our repo. It should be downloaded from a release instead.

https://github.com/webrecorder/wombat/pull/128 tracks our patch ; if this can't make it upstream, we should patch the downloaded version.

mgautierfr:
Fix #143


benoit74 (approved):
LGTM, I forced pushed to update the base commit following previous merge.

As discussed, this will be enhanced with the use of `hatch-openzim` plugin but this is sufficient for now to get rid of the external dependency from this repo.

Issue #149, opened 2024-01-17
Zimit2: Youtube videos are not playing (found at least in mesquartierschinois and 100rabbits)


benoit74:
Page: https://dev.library.kiwix.org/viewer#mes-quartiers-chinois_fr_all_2024-01/mesquartierschinois.wordpress.com/2014/07/17/girona-menamora/

When you try to play the youtube video on this page, the video never starts.

![image](https://github.com/openzim/warc2zim/assets/7102089/09503f3f-44c1-48b2-9ac0-4bf388ab4647)

mgautierfr:
This is strange. The video exists in the zim file (`youtube.fuzzy.replayweb.page/videoplayback%3Fid%3Do-AMT1WhCGLQGL0y0nkhuLvbKBBeOB1DzqsIIGuy4k6cYC`) and is loaded by the page/js.
But we have a (recurrent) request to `/www.youtube.com/api/stats/qoe%3F[lots_of_encoded_data]` which is 404.

I'm investigating.

benoit74:
Nota: I've updated the recipe `exclude` setting to `(\/\?|\/\#|\/wp-admin)` so that we scrape less way dummy things.

Jaifroid:
I corroborate this -- I've seen it often while testing support for zimit2 in Kiwix JS and Kiwix PWA (that video is usually not working). A small example is in the 100 rabbits zimit2 file from 2024-01, article "Orca". I'm seeing a bunch of POST failures in console log on Kiwix PWA (see bottom right of screenshot below). I don't know if that's related.

There is a December version of the same archive using zimit classic, where the video plays fine (in the PWA or KJS Browser Extension). These two small ZIMs ("Orca" article) might make a useful test case for comparison.

The main difference is that when running in zimit classic, there are console messages saying "Attempting to resolve fuzzy headers for url...", whereas I don't see any such message in the zimit2 implementation.

In my old implementation where I did the fuzzy search in the reader, it tells me that it transformed the requested URL (in the zimit classic version) into this monster in order to get the MP4:

`TRANSFORMED VIDEO URL C/A/youtube.com/embed/hQXa6TkSeH0 --> 
[...]

PR #158, opened 2024-01-25
Introduce generic Rewriter dispatching between specific rewriters.

mgautierfr:
Some js coming from specific website need to be patched to work "offline".

This patching is made using a set of specific regex to "activate" depending of path.

Sadly, there is no test because there is none in wabac.js (the source of the rules set) and
I don't know want/how to test.

We probably have more to import from wabac.js, but this PR is a fix to #149

### DIFF
--- a/src/warc2zim/content_rewriting/ds.py
+++ b/src/warc2zim/content_rewriting/ds.py
@@ -17,6 +17,7 @@ from warc2zim.content_rewriting.js import (
     add_prefix,
     m2str,
 )
+from warc2zim.content_rewriting.rx_replacer import add_suffix, replace_all
 
 
 @dataclass
@@ -29,22 +30,6 @@ def add_around(prefix: str, suffix: str) -> TransformationAction:
     @m2str
     def f(x) -> str:
         return prefix + x + suffix
-
-    return f
-
-
-def replace_all(text: str) -> TransformationAction:
-    @m2str
-    def f(_x) -> str:
-        return text
-
-    return f
-
-
-def add_suffix_all(suffix) -> TransformationAction:
-    @m2str
-    def f(x) -> str:
-        return x + suffix
 
     return f
 
@@ -196,19 +181,19 @@ RULES = [
             ),
             (
                 re.compile(r"yt\.setConfig.*PLAYER_CONFIG.*args\":\s*{"),
-                add_suffix_all(' "dash": "0", dashmpd: "", '),
+                add_suffix(' "dash": "0", dashmpd: "", '),
             ),
             (
                 re.compile(r"(?:\"player\":|ytplayer\.config).*\"args\":\s*{"),
-                add_suffix_all('"dash":"0","dashmpd":"",'),
+                add_suffix('"dash":"0","dashmpd":"",'),
             ),
             (
                 re.compile(r"yt\.setConfig.*PLAYER_VARS.*?{"),
-                add_suffix_all('"dash":"0","dashmpd":"",'),
+                add_suffix('"dash":"0","dashmpd":"",'),
             ),
             (
                 re.compile(r"ytplayer.config={args:\s*{"),
-                add_suffix_all('"dash":"0","dashmpd":"",'),
+                add_suffix('"dash":"0","dashmpd":"",'),
             ),
             (re.compile(r"\"0\"\s*?==\s*?\w+\.dash&&", re.M), replace_all("1&&")),
         ],
--- a/src/warc2zim/content_rewriting/ds.py
+++ b/src/warc2zim/content_rewriting/ds.py
@@ -17,21 +17,13 @@ from warc2zim.content_rewriting.js import (
     add_prefix,
     m2str,
 )
-from warc2zim.content_rewriting.rx_replacer import add_suffix, replace_all
+from warc2zim.content_rewriting.rx_replacer import add_around, add_suffix, replace_all
 
 
 @dataclass
 class SpecificRules:
     contains: list[str]
     rx_rules: list[TransformationRule]
-
-
-def add_around(prefix: str, suffix: str) -> TransformationAction:
-    @m2str
-    def f(x) -> str:
-        return prefix + x + suffix
-
-    return f
 
 
 MAX_BITRATE = 5000000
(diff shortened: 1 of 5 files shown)


rgaudin:
Is this a reimplementation of the JS version in wabac.js? I suppose it was not possible to source it directly.
What's the update process? I think we at least need a comment referencing the source and how to convert from it

mgautierfr:
> Is this a reimplementation of the JS version in wabac.js? I suppose it was not possible to source it directly.

Yes, it is.

> What's the update process? I think we at least need a comment referencing the source and how to convert from it

There is no update process. The simple things have been rewritten by chat-gpt 🙈 (and reviewed by a human). The other have been made by a human directly.

We could create a script which update the rules from the wabac source but it would not be straightforward as the "action" part of the rule is a bit different.
And if the code change, we will have to adapt manual anyway.


---

Last 4 commits are structural improvement but no functional changes.

benoit74:
Is it correct to add a comment in `ds.py` saying that all rules present in this file comes from `https://github.com/webrecorder/wabac.js/commits/main/src/rewrite/dsruleset.js`and have been manually converted from Javascript to Python (I would tell details about ChatGPT, who cares ...)? I would add this at least, as suggest by Renaud. I don't believe we can have a more precise conversion tool / procedure (unfortunately). 

I'm not very comfortable with the rest of the code, but "that's life" 😄

rgaudin:
> Is it correct to add a comment in `ds.py` saying that all rules present in this file comes from `https://github.com/webrecorder/wabac.js/commits/main/src/rewrite/dsruleset.js`and have been manually converted from Javascript to Python (I would tell details about ChatGPT, who cares ...)? I would add this at least, as suggest by Renaud. I don't believe we can have a more precise conversion tool / procedure (unfortunately).

That's what I have in mind ; just a starting point for another maintainer in the future: knowing where it came from, being able to check if there's a conversion bug and being able to convert new ones

benoit74:
Maybe also specify the last commit (from wabac.js) which has been taken into account, so that it is clear what is left to be back-ported

benoit74 (changes requested):
could you please change your PR to target `warc2zim2`, rebase your branch and resolve merge conflict (I thought it would be easier this way, also in order to close the other PR) and add the comment we discussed explaining where this code comes from?

benoit74:
Edit: github already decided to target `warc2zim2` automatically (nice ^^)

mgautierfr:
> add a comment in ds.py saying that all rules present in this file comes from https://github.com/webrecorder/wabac.js/commits/main/src/rewrite/dsruleset.jsand have been manually converted from Javascript to Python (I would tell details about ChatGPT, who cares ...)? 

There was already a comment in the commit message, but now it is also in the code.
(Directly in modifying the commit, no fixup)

src/warc2zim/content_rewriting/generic.py
```diff
@@ -0,0 +1,84 @@
...
+        if self.rewrite_mode == "javascript":
+            return self.rewrite_js(opts)
+
+        return ("", self.content)
+
+    def get_rewrite_mode(self, record, mimetype):
+        if getattr(record, "method", "GET") == "POST":
+            return None
+        if mimetype == "text/html":
+            # TODO : Handle header "Accept" == "application/json"
```
benoit74, line 50: What does it mean? Do you plan to do it soon? Is there an open issue around what has to be done?

src/warc2zim/content_rewriting/ds.py
```diff
@@ -0,0 +1,269 @@
...
+    rx_rules: list[TransformationRule]
+
+
+MAX_BITRATE = 5000000
+
+
+def set_max_bitrate(_opts: dict) -> int:
+    max_bitrate = MAX_BITRATE
+
+    # Extra opts is used in wabac, but we never set it ourself here.
```
benoit74, line 35: I'm sorry but I don't get what you mean here. Where do these `extraOpts` come from? The Twitter/Vimeo response? Then why don't we process them just like wabac does?

Maybe you should simply open an issue explaining:
- why we do not set extra_opts (we never did?)
- what could be the driver to decide to set them

src/warc2zim/content_rewriting/ds.py
```diff
@@ -0,0 +1,269 @@
+## THis file is almost a plain translation from js file
```
benoit74, line 1: Many typos / wrong English. Corrected version suggestion:
```
## This file is almost a plain translation from a JavaScript file.
## https://github.com/webrecorder/wabac.js/blob/219830cea0a732bdd72ce100fcfd2f3a1c9c3607/src/rewrite/dsruleset.js
## Partly translated by a human, partly by ChatGPT.

## The ruleset should be kept in sync with the original ones, but there is no easy
## automatic process for that at the moment.
```

benoit74 (changes requested):
Almost OK, two important questions (could be tackled in a separate issue if needed) and a suggestion to fix typos.

src/warc2zim/content_rewriting/generic.py
mgautierfr: Wabac is a checking if the header `Accept` is `application/json` and then use the json converter even if mimetype is `text/html`.

We don't have a json rewriter (we probably need one) and so I've just left a comment here to keep a trace that we may have something to do here.

> Do you plan to do it soon?

This has to be discussed. Until now, we had implement transformation based on what needed (this is why we agreed to not keep response header for now). But we may change our strategy and reimplement (blindly) the transformation in wabac. I will open a issue.

src/warc2zim/content_rewriting/ds.py
mgautierfr: We are two :)

Wabac's `set_max_bitrate` is searching/saving data in `extraOpts`. But I don't get from where it comes from and what to do with it.

src/warc2zim/content_rewriting/ds.py
mgautierfr: Fixed (directly in the commit)

src/warc2zim/content_rewriting/generic.py
benoit74: I've open an issue to track the question (at least): https://github.com/openzim/warc2zim/issues/160

src/warc2zim/content_rewriting/ds.py
benoit74: I've opened https://github.com/openzim/warc2zim/issues/161 to track the question (at least)

PR #168, opened 2024-01-31
Allow to specify a scraper suffix for the ZIM scraper metadata at the CLI

### LINKED ISSUE #156: Allow to specify the ZIM scraper metadata at the CLI
Currently, warc2zim sets the ZIM scraper metadata as `warc2zim x.y.z` (where `x.y.z` is warc2zim version).

It would be beneficial if we could instead specify the scraper at the CLI for cases where warc2zim is only used as a dependency in another scraper, since it is not totally correct to say that warc2zim who is the scraper.

E.g. when it is zimit who is "using" warczim, we would pass `zimit u.v.w` as `--scraper` and the final ZIM scraper metadata would become `zimit u.v.w`

`warc2zim` version won't be obvious anymore in the ZIM, but just like any other scraper dependency. But it is still possible to find `warc2zim` version based on `zimit` version, since it is either explicit in Zimit changelog or at least in Zimit git history.

benoit74:
Fix #156 

Nota: also updates the version to `dev1` as agreed in issue.

### DIFF
--- a/src/warc2zim/__about__.py
+++ b/src/warc2zim/__about__.py
@@ -1,1 +1,1 @@
-__version__ = "2.0.0-dev0"
+__version__ = "2.0.0-dev1"
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -142,6 +142,8 @@ class Converter:
 
         self.written_records = self.total_records = 0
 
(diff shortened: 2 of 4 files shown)


rgaudin (approved):
LGTM

PR #169, opened 2024-01-31
Enhance wording about return code

benoit74:
Logs were speaking about `error code 100`, which indicates that args are OK for warc2zim.

I changed this to `return code 100` since it is the proper term and the wrong term was frequently misleading the content team into believing there was an error.


rgaudin (approved):
LGTM

PR #170, opened 2024-01-31
String leading and trailing space from url.

### LINKED ISSUE #167: Zimit2 lowtechmagazin scrape seems to be missing content
See:
* https://dev.library.kiwix.org/viewer#low-tech-solutions/
* https://dev.library.kiwix.org/viewer#offline-reading/
* https://dev.library.kiwix.org/viewer#donate/

All available links at https://dev.library.kiwix.org/viewer#solar.lowtechmagazine.com_en_all_2024-01/solar.lowtechmagazine.com/

mgautierfr:
Fix #167

Spec is explicit about that, we can have space around url: https://www.w3.org/TR/2014/REC-html5-20141028/infrastructure.html#valid-url-potentially-surrounded-by-spaces

### DIFF
--- a/src/warc2zim/url_rewriting.py
+++ b/src/warc2zim/url_rewriting.py
@@ -170,6 +170,8 @@ class ArticleUrlRewriter:
         The url is "fully" rewrited to point to a normalized entry path
         """
 
+        url = url.strip()
+
         if url.startswith("data:") or url.startswith("blob:"):
             return url
 
--- a/tests/test_url_rewriting.py
+++ b/tests/test_url_rewriting.py
@@ -36,12 +36,13 @@ def test_relative_url(rewriter):
 
 def test_absolute_path_url(rewriter):
     for url in ["/foo", "/foo/bar"]:
-        rewriten = rewriter(url)
-        # Must produce a relative link.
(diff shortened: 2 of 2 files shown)


benoit74 (changes requested):
This is not sufficient to fix the issue.

When there is a space around a URL, the associated WARC item is now properly included inside the ZIM, this is OK. We can access it if we know its URL, this is OK.

But there is a second part (which was not yet clearly identified in the issue, I admit) linked to the fact that the URL is absolute in the source website, and stays absolute inside the ZIM, meaning it won't work on kiwix-serve (for instance).

You will find a small test WARC here (I used `--depth 1` so only home page and pages linked from the homepage will be available): https://tmp.kiwix.org/ci/test-warc/solar.lowtechmagazine.com_2024-02/solar.lowtechmagazine.com-depth1.warc.gz

(and BTW, I move bbc_persian test WARCs in same `ci/test-warc` parent folder

benoit74:
Oh, but it was already inside the ZIM ; then it is just not working in my test ... that's weird. Let me check again

benoit74 (approved):
Never mind about my previous remark, I ran improper code base ... 🤦🏻

LGTM

Issue #165, opened 2024-01-30
Zimit2: Vimeo videos are not working


Jaifroid:
On this page: https://dev.library.kiwix.org/content/mes-quartiers-chinois_fr_all_2024-01/mesquartierschinois.wordpress.com/2014/07/10/will-blog-for-croquetas/ there is a Vimeo-based video that is correctly scraped and is in the ZIM. However, it doesn't load from that page, with a somewhat cryptic message in wombat.js about XHR sync not supported in a ServiceWorker-based player (see badly drawn arrow in screenshot). I can't check whether Replay-based versions (zimit classic) of this ZIM could play this video because they won't load on the dev server.

This probably isn't a blocker, as only one Vimeo video is in this ZIM.

![image](https://github.com/openzim/warc2zim/assets/4304337/33d4a131-adb3-4b9f-8ad8-d5354be2c7f0)

Jaifroid:
Further info: as you can see below, the video is scraped and is in the ZIM, and plays if the reader can select it directly from its ZIM URL:

![image](https://github.com/openzim/warc2zim/assets/4304337/ae11260b-15bf-4774-98c4-d60e6601d9bf)

![image](https://github.com/openzim/warc2zim/assets/4304337/4983d7f9-5fca-46e3-977e-0cb0df042092)

Jaifroid:
And I've just noticed that there is a more significant error at the top of the console log. Seems like there is an issue loading `wombatWorkers.js` (this should be a Web Worker, not a Service Worker, so they should be loadable):

`Uncaught DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'https://dev.library.kiwix.org/content/mes-quartiers-chinois_fr_all_2024-01/_zim_static/wombatWorkers.js' failed to load.`

benoit74:
Thank you for reporting all this! I'm now convinced we need to setup a test website grouping all "weird" JS situations that are fixed by wabac DS rules at `src/warc2zim/content_rewriting/ds.py`

benoit74:
Closing in favor of #242, Vimeo videos support is limited but working now.

PR #172, opened 2024-02-01
Fix Vimeo video

mgautierfr:
This PR is **kind of** a fix to #165.

The root issue seems that vimeo is using js module (instead of plain "classic" script) and rewriter is not detecting thing correctly.

If it is somehow easily fixable for inline module, we have to detect that for js resources themselves. We have regex in JS rewriter to detect that, but it seems it is not working. The check
```python
if "module.js" in self.path:
    return "module"
else:
    return "javascript"
```
works in our use case, but probably (definitively?) not something we want in production.
I'm still searching for a good way to detect that.
For wabac.js, it is easier as js is rewritten as runtime, so it knows the context.

On top of that, the player seems to be broken sometime, depending if you are clicking on the "video background" or the "play" button. Alternate clicking on video/play button launch the video.


benoit74:
Superseeded by https://github.com/openzim/warc2zim/pull/228

PR #175, opened 2024-02-05
Be more tolerant to invalid css.

### LINKED ISSUE #155: Failure of `thalesdoc_en_all`: AttributeError: 'NoneType' object has no attribute 'startswith'
We have an error in zimit2 for https://farm.openzim.org/pipeline/2119ed82-41aa-4561-a359-8e75968b445c/debug

```
  File "/usr/bin/zimit", line 566, in <module>
    zimit()
  File "/usr/bin/zimit", line 464, in zimit
    return warc2zim(warc2zim_args)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/main.py", line 89, in main
    return converter.run()
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/converter.py", line 277, in run
    self.add_items_for_warc_record(record)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/converter.py", line 490, in add_items_for_warc_record
    payload_item = WARCPayloadItem(
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/items.py", line 72, in __init__
    ).rewrite(self.content)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 76, in rewrite
    self.feed(content)
  File "/usr/lib/python3.10/html/parser.py", line 110, in feed
    self.goahead(0)
  File "/usr/lib/python3.10/html/parser.py", line 170, in goahead
    k = self.parse_starttag(i)
  File "/usr/lib/python3.10/html/parser.py", line 344, in parse_starttag
    self.handle_starttag(tag, attrs)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 96, in handle_starttag
    self.send(transform_attrs(attrs, url_rewriter, self.css_rewriter))
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 46, in transform_attrs
    return " ".join(format_attr(*attr) for attr in processed_attrs)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 46, in <genexpr>
    return " ".join(format_attr(*attr) for attr in processed_attrs)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 45, in <genexpr>
    processed_attrs = (process_attr(attr, url_rewriter, css_rewriter) for attr in attrs)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 20, in process_attr
    return (attr[0], url_rewriter(attr[1]))
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/content_rewriting/html.py", line 93, in <lambda>
    url_rewriter = lambda url: self.url_rewriter(url, False)
  File "/app/zimit/lib/python3.10/site-packages/warc2zim/url_rewriting.py", line 161, in __call__
    if url.startswith("data:") or url.startswith("blob:"):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

Looks like we have one more `# pyright: ignore` which should not be there in my last PR (`url` can indeed by None and it is an issue).

```
        if tag == "a":
            url_rewriter = lambda url: self.url_rewriter(  # noqa: E731
                url, False  # pyright: ignore
            )
```

mgautierfr:
Return the original content if we cannot parse the css.

---

In case of invalid css tinycss2 doesn't not failed on parsing, but generate a ParserError token
https://doc.courtbouillon.org/tinycss2/stable/api_reference.html?highlight=parser#tinycss2.ast.ParseError

This is on serialization that exception occurs.
We even try/except all transformation to be sure.

Fix #155

### DIFF
--- a/src/warc2zim/content_rewriting/css.py
+++ b/src/warc2zim/content_rewriting/css.py
@@ -10,6 +10,7 @@ from tinycss2 import (
 )
 from tinycss2.serializer import serialize_url
 
+from warc2zim.constants import logger
 from warc2zim.content_rewriting import UrlRewriterProto
 from warc2zim.content_rewriting.rx_replacer import RxRewriter
 
@@ -51,6 +52,13 @@ class CssRewriter:
             # If tinycss fail to parse css, it will generate a "Error" token.
             # Exception is raised at serialization time.
             # We try/catch the whole process to be sure anyway.
+            logger.warning(
+                (
+                    "Css transformation fails. Fallback to regex rewriter.\n"
+                    "Article path is %s"
+                ),
+                self.url_rewriter.article_url,
+            )
             return self.fallback_rewriter.rewrite_content(content, {})
         return output
 
@@ -64,6 +72,13 @@ class CssRewriter:
             # If tinycss fail to parse css, it will generate a "Error" token.
             # Exception is raised at serialization time.
             # We try/catch the whole process to be sure anyway.
+            logger.warning(
+                (
+                    "Css transformation fails. Fallback to regex rewriter.\n"
+                    "Content is `%s`"
+                ),
+                content,
+            )
             return self.fallback_rewriter.rewrite_content(content, {})
 
     def process_list(self, components: Iterable[ast.Node]):
--- a/src/warc2zim/content_rewriting/css.py
+++ b/src/warc2zim/content_rewriting/css.py
@@ -1,3 +1,4 @@
+import re
 from collections.abc import Iterable
 
 from tinycss2 import (
@@ -10,27 +11,60 @@ from tinycss2 import (
 from tinycss2.serializer import serialize_url
 
 from warc2zim.content_rewriting import UrlRewriterProto
+from warc2zim.content_rewriting.rx_replacer import RxRewriter
+
+
+class FallbackRegexCssRewriter(RxRewriter):
+    def __init__(self, url_rewriter: UrlRewriterProto):
+        rules = [
+            (
+                re.compile(r"""url\((?P<quote>['"])?(?P<url>.+?)(?P=quote)(?<!\\)\)"""),
+                lambda m_object, _opts: "".join(
+                    [
+                        "url(",
+                        m_object["quote"],
+                        url_rewriter(m_object["url"]),
+                        m_object["quote"],
+                        ")",
+                    ]
+                ),
+            )
+        ]
+        super().__init__(rules)
 
 
 class CssRewriter:
     def __init__(self, url_rewriter: UrlRewriterProto):
         self.url_rewriter = url_rewriter
+        self.fallback_rewriter = FallbackRegexCssRewriter(url_rewriter)
 
     def rewrite(self, content: str | bytes) -> str:
-        if isinstance(content, bytes):
-            rules = parse_stylesheet_bytes(content)[0]
-        else:
-            rules = parse_stylesheet(content)
-        self.process_list(rules)
+        try:
+            if isinstance(content, bytes):
+                rules = parse_stylesheet_bytes(content)[0]
+            else:
+                rules = parse_stylesheet(content)
+            self.process_list(rules)
 
-        output = serialize(rules)
+            output = serialize(rules)
+        except Exception:
+            # If tinycss fail to parse css, it will generate a "Error" token.
+            # Exception is raised at serialization time.
+            # We try/catch the whole process to be sure anyway.
+            return self.fallback_rewriter.rewrite_content(content, {})
         return output
 
     def rewrite_inline(self, content: str) -> str:
-        rules = parse_declaration_list(content)
-        self.process_list(rules)
-        output = serialize(rules)
-        return output
+        try:
+            rules = parse_declaration_list(content)
+            self.process_list(rules)
+            output = serialize(rules)
+            return output
+        except Exception:
+            # If tinycss fail to parse css, it will generate a "Error" token.
+            # Exception is raised at serialization time.
+            # We try/catch the whole process to be sure anyway.
+            return self.fallback_rewriter.rewrite_content(content, {})
 
     def process_list(self, components: Iterable[ast.Node]):
         if components:  # May be null
--- a/tests/test_js_rewriting.py
+++ b/tests/test_js_rewriting.py
@@ -26,8 +26,8 @@ def rewrite_this_js_content(request):
 
 def test_this_js_rewrite(rewrite_this_js_content):
     assert (
-        JsRewriter(lambda x: x).rewrite(rewrite_this_js_content.input_)
-        == rewrite_this_js_content.expected
+        JsRewriter(lambda x: x).rewrite(rewrite_this_js_content.input_str)
+        == rewrite_this_js_content.expected_str
     )
 
 
@@ -65,7 +65,7 @@ class WrappedTestContent(ContentForTests):
 
     def __post_init__(self):
         super().__post_init__()
-        self.expected = self.wrap_script(self.expected)
+        self.expected = self.wrap_script(self.expected_str)
 
 
 @pytest.fixture(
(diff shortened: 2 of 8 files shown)


src/warc2zim/content_rewriting/css.py
```diff
@@ -23,14 +23,30 @@ def rewrite(self, content: str | bytes) -> str:
...
         self.process_list(rules)
 
-        output = serialize(rules)
+        try:
+            output = serialize(rules)
+        except Exception:
+            import sys  # Import sys only in case of error
+
+            # If we cannot parse the css, simply return the original content
+            print("Error rewriting CSS", file=sys.stderr)  # noqa
```
benoit74, line 32: Please use a logger instead of print. 

It looks like it has not yet been used in this file, so you will have to `from warc2zim.constants import logger` first.

I suggest these logs are written only at "DEBUG" log level.

src/warc2zim/content_rewriting/css.py
```diff
@@ -23,14 +23,30 @@ def rewrite(self, content: str | bytes) -> str:
             rules = parse_stylesheet(content)
         self.process_list(rules)
 
-        output = serialize(rules)
+        try:
+            output = serialize(rules)
+        except Exception:
+            import sys  # Import sys only in case of error
```
benoit74, line 29: I'm a real beginner at this, but does it really make much of a difference (compared to importing once at the module level)? I find it pretty weird (especially since you have to add a comment)

benoit74 (changes requested):
See comments

mgautierfr:
New version rewrite the css content using regex if tinycss2 fails to parse it.

src/warc2zim/content_rewriting/css.py
```diff
@@ -23,14 +36,26 @@ def rewrite(self, content: str | bytes) -> str:
...
 
-        output = serialize(rules)
+        try:
+            output = serialize(rules)
+        except Exception:
+            fallback_rewriter = FallbackRegexCssRewriter(self.url_rewriter)
+            if isinstance(content, bytes):
+                content = content.decode()
+            return fallback_rewriter.rewrite_content(
+                content, {}  # pyright: ignore[reportArgumentType]
```
benoit74, line 46: Why do we need this pyright rule? I checked but I don't understand why we have an error and need this rule (maybe add a comment to explain it so that it is clearer for next maintainer)

tests/test_js_rewriting.py
```diff
@@ -65,7 +65,9 @@ def wrap_script(text: str) -> str:
 
     def __post_init__(self):
         super().__post_init__()
-        self.expected = self.wrap_script(self.expected)
+        self.expected = self.wrap_script(
+            self.expected  # pyright: ignore[reportArgumentType]
```
benoit74, line 69: Please add a getter method (something like `expected_str`) in ContentForTest which raises an error if `expected` is not a string and return `expected` otherwise ; and use it here. Otherwise the tests will begin to be full of `pyright: ignore`

tests/test_js_rewriting.py
```diff
@@ -131,7 +133,9 @@ def wrap_import(text: str) -> str:
     def __post_init__(self):
         super().__post_init__()
         self.article_url = "https://exemple.com/some/path/"
-        self.expected = self.wrap_import(self.expected)
+        self.expected = self.wrap_import(
+            self.expected  # pyright: ignore[reportArgumentType]
```
benoit74, line 137: Use it here as well

src/warc2zim/content_rewriting/css.py
```diff
@@ -23,14 +36,26 @@ def rewrite(self, content: str | bytes) -> str:
             rules = parse_stylesheet(content)
         self.process_list(rules)
 
-        output = serialize(rules)
+        try:
+            output = serialize(rules)
+        except Exception:
+            fallback_rewriter = FallbackRegexCssRewriter(self.url_rewriter)
```
benoit74, line 42: I would prefer to initialize the `fallback_rewriter` in the `__init__`, it feels weird to have two places where we initialize this object (inline and not inline) and potentially two different ways to do it. WDYT?

src/warc2zim/content_rewriting/css.py
```diff
@@ -23,14 +36,26 @@ def rewrite(self, content: str | bytes) -> str:
...
+            if isinstance(content, bytes):
```
benoit74, line 43: Is it possible to move this inside the generic RxRewriter so that it could benefit other rewriters? (I don't expect we will have many ways to convert a `bytes` into a `str` depending on the rewriter)

src/warc2zim/content_rewriting/css.py
```diff
@@ -23,14 +36,26 @@ def rewrite(self, content: str | bytes) -> str:
             rules = parse_stylesheet(content)
         self.process_list(rules)
 
-        output = serialize(rules)
+        try:
```
benoit74, line 39: Why are we catching exception only in serialization and not the whole processing? At least it is not homogeneous with the code for inline CSS processing, which makes it weird and at least deserve a comment explaining the reason.

benoit74 (changes requested):
See comments.

Please add more tests to have 100% test coverage on the patch : https://app.codecov.io/gh/openzim/warc2zim/pull/175?src=pr&el=tree ; no need to do all test cases on all code branches, but at least have one test case on each of them, especially since here you are not testing the inline situation which is the one we have in production.

I think that we need:
- logs when the normal CSS rewriting failed (a warning probably)
- exception handling for the fallback rewriter: maybe simply a clear log with original content, because this is probably a very rare situation (I don't see why regex would fail) but it would make debugging way simpler
- more fine-grained processing of CSS rules ; since we know that tinycss is permissive in terms of parsing, we could process / serialize CSS rules one by one and apply the fallback only on failing CSS rules (this could be done in a separate issue)

src/warc2zim/content_rewriting/css.py
mgautierfr: This is because pyright "promotes" `bytes` type to `bytes|memoryview`.
See the first commit.

Maybe it could be backported in our hatch config.

benoit74: Great! I approve your solution, see https://github.com/openzim/_python-bootstrap/issues/37

benoit74 (changes requested):
All conversations have been addressed properly, thank you (and great!), but AFAIK you didn't provided feedback on the last 3 recommendations of https://github.com/openzim/warc2zim/pull/175#pullrequestreview-1864515066

mgautierfr:
Sorry, I forget about these comments

> logs when the normal CSS rewriting failed (a warning probably)

Indeed, will do

> exception handling for the fallback rewriter: maybe simply a clear log with original content, because this is probably a very rare situation (I don't see why regex would fail) but it would make debugging way simpler

As you say, I don't see why regex would fail. If it is the case, it is more a bug in our code than a invalid input we have to handle somehow.

> more fine-grained processing of CSS rules ; since we know that tinycss is permissive in terms of parsing, we could process / serialize CSS rules one by one and apply the fallback only on failing CSS rules (this could be done in a separate issue)

I have think about that, but not an easy task. We would have to reimplement tinycss2 serialize feature. And on top of that, it is not clear about what to do in case of parsing error. Even tinycss try to be smart and recover error but give up on parsing error (https://github.com/Kozea/tinycss2/blob/main/tinycss2/ast.py#L105-L115).

benoit74:
> As you say, I don't see why regex would fail. If it is the case, it is more a bug in our code than a invalid input we have to handle somehow.

OK, let's live with it for now.

> I have think about that, but not an easy task. We would have to reimplement tinycss2 serialize feature. And on top of that, it is not clear about what to do in case of parsing error. Even tinycss try to be smart and recover error but give up on parsing error ([Kozea/tinycss2@main/tinycss2/ast.py#L105-L115](https://github.com/Kozea/tinycss2/blob/main/tinycss2/ast.py?rgh-link-date=2024-02-08T10%3A32%3A05Z#L105-L115)).

Agreed! I've opened an issue to keep this information in mind: https://github.com/openzim/warc2zim/issues/176

benoit74 (approved):
LGTM, I will merge it

PR #183, opened 2024-02-13
Correctly detect encoding and decode bytes 

### LINKED ISSUE #177: Failure of `thalesdoc`: invalid UTF-8 start byte in an HTML file
With latest zimit2 image (including with warc2zim 2.0.0-dev3), we face a new issue (I processed locally the full crawl I recently uploaded at https://tmp.kiwix.org/ci/test-warc/thalesdoc_en_all_full_2024-02-04/) around invalid UTF-8 in an HTML file.

Last log lines:
```
[warc2zim::2024-02-09 10:44:49,443] WARNING:Css transformation fails. Fallback to regex rewriter.
Content is `border-left-style: solid;border-left-width: 1px;border-left-color: #c0c0c0;border-right-style: solid;border-right-width: 1px;border-right-color: #c0c0c0;border-top-style: solid;border-top-width: 1px;border-top-color: #c0c0c0;border-bottom-style: solid;border-bottom-width: 1px;border-bottom-color: #c0c0c0;w`
[warc2zim::2024-02-09 10:45:10,064] WARNING:Css transformation fails. Fallback to regex rewriter.
Content is `width: 100%; cellspacing=`
  File "<stdin>", line 1, in <module>
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/main.py", line 90, in main
    return converter.run()
           ^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/converter.py", line 264, in run
    self.add_items_for_warc_record(record)
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/converter.py", line 493, in add_items_for_warc_record
    payload_item = WARCPayloadItem(
                   ^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/items.py", line 38, in __init__
    (self.title, self.content) = Rewriter(path, record, known_urls).rewrite(
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/content_rewriting/generic.py", line 36, in rewrite
    return self.rewrite_html(head_template, css_insert)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/content_rewriting/generic.py", line 72, in rewrite_html
    return HtmlRewriter(self.url_rewriter, head_insert, css_insert).rewrite(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/content_rewriting/html.py", line 79, in rewrite
    content = to_string(content)
              ^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/utils.py", line 47, in to_string
    input_ = input_.decode(  # pyright: ignore[reportGeneralTypeIssues, reportAttributeAccessIssue]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/encodings/utf_8_sig.py", line 23, in decode
    (output, consumed) = codecs.utf_8_decode(input, errors, True)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x96 in position 6224: invalid start byte
```

mgautierfr:
This PR try to detect encoding using 3 methods:
- From http headers in the warc record
- From encoding declaration inside the content
- From statistical analysis of the content (made by chardet)

Fix #177

### DIFF
--- a/src/warc2zim/utils.py
+++ b/src/warc2zim/utils.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import re
 
+import chardet
 from bs4 import BeautifulSoup
 from warcio.recordloader import ArcWarcRecord
 
@@ -57,14 +58,38 @@ def get_record_encoding(record: ArcWarcRecord) -> str | None:
     if m := ENCODING_RE.search(content_type):
         return m.group("encoding")
 
-def to_string(input_: str | bytes) -> str:
-    try:
-        input_ = input_.decode(  # pyright: ignore[reportGeneralTypeIssues, reportAttributeAccessIssue]
-            "utf-8-sig"
-        )
-    except AttributeError:
-        pass
-    return input_  # pyright: ignore[reportGeneralTypeIssues, reportReturnType]
+
+def to_string(input_: str | bytes, encoding: str | None) -> str:
+    """
+    Decode content to string, trying to be the more tolerant possible to invalid
+    declared encoding.
+
+    This try decode the content using 3 methods:
+     - From http headers in the warc record (given as `encoding` argument)
+     - From encoding declaration inside the content (hopping that content can be
+       losely decode using ascii to something usable)
+     - From statistical analysis of the content (made by chardet)
+
+    """
+    if isinstance(input_, str):
+        return input_
+
+    if encoding:
+        try:
+            return input_.decode(encoding)
+        except ValueError:
+            pass
+
+    # Detect encoding from content.
+    content_start = input_[:1024].decode("ascii", errors="replace")
+    if m := ENCODING_RE.search(content_start):
+        encoding = m.group("encoding")
+    else:
+        encoding = chardet.detect(input_)["encoding"]
+
+    if not encoding:
+        raise ValueError(f"Impossible to detect encoding of content {input_[:200]}")
+    return input_.decode(encoding)
 
 
 def get_record_content(record: ArcWarcRecord):
--- a/src/warc2zim/utils.py
+++ b/src/warc2zim/utils.py
@@ -3,10 +3,17 @@
 
 from __future__ import annotations
 
+import re
+
 from bs4 import BeautifulSoup
 from warcio.recordloader import ArcWarcRecord
 
 from warc2zim.__about__ import __version__
+
+ENCODING_RE = re.compile(
+    r"(charset|encoding)=(?P<quote>['\"]?)(?P<encoding>[a-wA-Z0-9_\-]+)(?P=quote)",
+    re.ASCII,
+)
 
 
 def get_version():
@@ -44,6 +51,11 @@ def parse_title(content):
     except Exception:
         return ""
 
+
+def get_record_encoding(record: ArcWarcRecord) -> str | None:
+    content_type = get_record_content_type(record)
+    if m := ENCODING_RE.search(content_type):
+        return m.group("encoding")
 
 def to_string(input_: str | bytes) -> str:
     try:
--- a/src/warc2zim/utils.py
+++ b/src/warc2zim/utils.py
@@ -87,13 +87,16 @@ def to_string(input_: str | bytes, encoding: str | None) -> str:
     # Detect encoding from content.
     content_start = input_[:1024].decode("ascii", errors="replace")
     if m := ENCODING_RE.search(content_start):
-        encoding = m.group("encoding")
+        encodings = [m.group("encoding")]
     else:
-        encoding = chardet.detect(input_)["encoding"]
+        encodings = [e["encoding"] for e in chardet.detect_all(input_)]
 
-    if not encoding:
-        raise ValueError(f"Impossible to detect encoding of content {input_[:200]}")
-    return input_.decode(encoding)
+    for encoding in encodings:
+        try:
+            return input_.decode(encoding)
+        except ValueError:
+            pass
+    raise ValueError(f"Impossible to decode content {input_[:200]}")
 
 
 def get_record_content(record: ArcWarcRecord):
--- a/src/warc2zim/utils.py
+++ b/src/warc2zim/utils.py
@@ -71,6 +71,7 @@ def to_string(input_: str | bytes, encoding: str | None) -> str:
      - From statistical analysis of the content (made by chardet)
 
     """
+    tried_encodings = set()
     if isinstance(input_, str):
         return input_
 
@@ -82,20 +83,24 @@ def to_string(input_: str | bytes, encoding: str | None) -> str:
         try:
             return input_.decode(encoding)
         except ValueError:
+            tried_encodings.add(encoding)
             pass
 
     # Detect encoding from content.
     content_start = input_[:1024].decode("ascii", errors="replace")
     if m := ENCODING_RE.search(content_start):
         encoding = m.group("encoding")
-        if encoding:
+        if encoding and encoding not in tried_encodings:
             try:
                 return input_.decode(encoding)
             except ValueError:
+                tried_encodings.add(encoding)
                 pass
 
     encodings = (
-        encoding for e in chardet.detect_all(input_) if (encoding := e["encoding"])
+        encoding
+        for e in chardet.detect_all(input_)
+        if (encoding := e["encoding"]) and encoding not in tried_encodings
     )
 
     for encoding in encodings:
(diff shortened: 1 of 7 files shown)


src/warc2zim/utils.py
```diff
@@ -42,14 +53,43 @@ def parse_title(content):
...
-    try:
-        input_ = input_.decode(  # pyright: ignore[reportGeneralTypeIssues, reportAttributeAccessIssue]
-            "utf-8-sig"
-        )
-    except AttributeError:
-        pass
-    return input_  # pyright: ignore[reportGeneralTypeIssues, reportReturnType]
+def get_record_encoding(record: ArcWarcRecord) -> str | None:
+    content_type = get_record_content_type(record)
+    if m := ENCODING_RE.search(content_type):
```
rgaudin, line 58: This will never work. that regexp looks for `encoding=xxx` pattern while your content-type is _just_ the value from the headers

rgaudin (changes requested):
It clearly lacks tests. We know this is a fragile area. Maybe we should add a test warc to ensure we're testing the overall process.

rgaudin:
Ah and I forgot the reason I looked at this 😅 The PR does not explains what we're trying to fix here. The ticket is just a report of warc2zim crashing with no answer.

Why is it failing? I guess we need to decode to rewrite it. Are we storing the result in original encoding or UTF-8 as ZIM spec requests?

src/warc2zim/utils.py
mgautierfr: Regex searches for "charset=xxx" or "encoding=xxx" and if content-type is indeed just the value from the headers, this is the way to declare encoding in the header:
https://www.w3.org/International/articles/http-charset/index

mgautierfr:
> I think it would make more sense to:
>- first try the encoding of the Warc Record
>- if this fails or there is none, then try all the encodings found in the file
>- if this fails or there is none, then try all the encodings detected by chardet

[Specification](https://www.w3.org/TR/html4/charset.html#h-5.2.2) says otherwise. We should first use the encoding declared in the header, then the one in the meta tag then heuristic.

src/warc2zim/utils.py
rgaudin: My bad 👍

mgautierfr:
> Ah and I forgot the reason I looked at this 😅 The PR does not explains what we're trying to fix here. The ticket is just a report of warc2zim crashing with no answer.
> 
> Why is it failing? I guess we need to decode to rewrite it. Are we storing the result in original encoding or UTF-8 as ZIM spec requests?

I have put a comment in the issue. The problem is that input content may not be encoded in utf8. We have to be "smart" to select the right encoding.

We still use utf8 encoding when we save the content in the zim file. This PR is just about decoding content in the warc.

src/warc2zim/utils.py
```diff
@@ -42,14 +53,53 @@ def parse_title(content):
...
+        return m.group("encoding")
+
+
+def to_string(input_: str | bytes, encoding: str | None) -> str:
+    tried_encodings = set()
+    if isinstance(input_, str):
+        return input_
```
benoit74, line 65: This should be covered as well, it is straightforward.

src/warc2zim/utils.py
```diff
@@ -42,14 +53,53 @@ def parse_title(content):
...
+
+    if encoding:
+        try:
+            return input_.decode(encoding)
+        except (ValueError, LookupError):
+            tried_encodings.add(encoding)
+            pass
+
+    # Detect encoding from content.
+    # First try declared encoding in content.
```
benoit74, line 79: If you add a comment stating "First try declared encoding in content", you should add another comment stating "Then try ..."

And I would rather change this to the list you mentioned in your PR (first use the HTTP headers, then ..., then ...) with a link to HTML specification you mentioned.

This should probably be placed in the docstring of the method in fact.

src/warc2zim/utils.py
```diff
@@ -42,14 +53,53 @@ def parse_title(content):
...
+                return input_.decode(encoding)
+            except (ValueError, LookupError):
+                tried_encodings.add(encoding)
+                pass
+
+    all_encodings = chardet.detect_all(input_)
+    encodings = [e["encoding"] for e in all_encodings if e["encoding"]]
+
+    if not encodings:
+        raise ValueError(f"Impossible to detect encoding of content {input_[:200]}")
```
benoit74, line 94: Either cover this line or add a pragma to indicate this is not feasible

src/warc2zim/utils.py
```diff
@@ -42,14 +53,53 @@ def parse_title(content):
...
+    for encoding in encodings:
+        if encoding in tried_encodings:
+            continue
```
benoit74, line 97: This should be covered, pretty easy

src/warc2zim/utils.py
```diff
@@ -42,14 +53,53 @@ def parse_title(content):
...
+        try:
+            return input_.decode(encoding)
+        except ValueError:
+            pass
+    raise ValueError(f"Impossible to decode content {input_[:200]}")
```
benoit74, line 102: This should be covered as well, this is what we are arguing about whether we keep this exception and let the scraper fail or if we enhance the logic.

benoit74 (changes requested):
Thank you for these changes, we are moving in the right direction.

We still need few more tests (especially the one which makes the to_string and the scraper crash due to "mixed-encoding" in the file, you might reuse the content I built for the test website)

src/warc2zim/utils.py
mgautierfr: Added a docstring.

src/warc2zim/utils.py
mgautierfr: Removed. I never succeed to have this condition, and it is "covered" by the loop just after.

src/warc2zim/utils.py
mgautierfr: Test added.

benoit74 (approved):
LGTM, @rgaudin could you make a second pass please?

src/warc2zim/utils.py
```diff
@@ -42,14 +53,61 @@ def parse_title(content):
...
+            return input_.decode(encoding)
+        except (ValueError, LookupError):
+            tried_encodings.add(encoding)
+            pass
+
+    # Detect encoding from content.
+    content_start = input_[:1024].decode("ascii", errors="replace")
+    if m := ENCODING_RE.search(content_start):
+        encoding = m.group("encoding")
+        if encoding and encoding not in tried_encodings:
```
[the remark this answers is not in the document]
mgautierfr: I haven't put a pragma here because it is not something that we shouldn't test.
We should but it kinda needs some complex test and we don't really care. But it is a real feature with a code path production code can takes.

It is not the same as test for debugging purpose or really rare conditions.

We are not testing 100% of the patch, that's a reality. Let's not mask it.
(But I will not argue more, if you want it, you will have it)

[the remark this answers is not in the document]
benoit74: I'm fine with it (or without it more exactly) 🤣

src/warc2zim/utils.py
```diff
@@ -42,14 +53,61 @@ def parse_title(content):
...
+            try:
+                return input_.decode(encoding)
+            except (ValueError, LookupError):
+                tried_encodings.add(encoding)
+                pass
+
+    all_encodings = chardet.detect_all(input_)
```
rgaudin, line 100:
```suggestion
    encodings = [e["encoding"] for e in chardet.detect_all(input_) if e["encoding"]]
```
and skip the next line

You can even save the continue block later with

```py
encodings = [e["encoding"] for e in chardet.detect_all(input_) if e["encoding"] and e["encoding"] not in tried_encodings]]
```

rgaudin (approved):
LGTM ; a simple format suggestion

src/warc2zim/utils.py
mgautierfr: Done (Directly in commits)

PR #184, opened 2024-02-14
Add first version of a test website

### LINKED ISSUE #166: Setup a test website with all weird JS situations
See https://github.com/openzim/warc2zim/issues/165#issuecomment-1918546746

We need to setup a test website (and associated Zimfarm recipe) grouping all "weird" JS situations that are fixed by wabac DS rules at `src/warc2zim/content_rewriting/ds.py`, so that it will be quick to test all of them at once.

benoit74:
Fix #166 (and beyond)

## Changes
- Create a test website with many weird content that has to be processed by zimit/warc2zim


benoit74:
This PR is obviously only a first step forward, many other cases have to be generated.

I pushed the resulting test website manually to https://tmp.kiwix.org/ci/test-website/

I intend to automatize publishing via a Github Action once you confirm it is a good idea to publish there.

test-website/bad-encoding.html
```diff
@@ -0,0 +1,31 @@
...
+
+<body>
+
+    <h2>Bad encoding</h2>
+
+    <p>
+        <img src="http://www.mediaelementjs.com/images/big_buck_bunny.jpg">
+    </p>
+
+    <p>Below is some bad encoding. This file has been saved with Windows 1252 and content below comes from Simplified Chinese GB2312. Typically happens when content is merged from various sources, e.g. in dynamic websites.</p>
```
mgautierfr, line 23: Maybe add in the description that html meta declare encoding as utf-8 (and that firefox decode this content as utf-8 "by default")

test-website/chinese-encoding.html
```diff
@@ -0,0 +1,27 @@
...
+
+<body>
+
+    <h2>Chinese encoding</h2>
+
+    <p>
+        <img src="http://www.mediaelementjs.com/images/big_buck_bunny.jpg">
+    </p>
+
+    <p>This file has been encoded with Simplified Chinese gb2312 encoding and simplified chinese characters: ����.</p>
```
mgautierfr, line 23: And firefox decode it as `GBK`...

mgautierfr (commented):
Just few comments. But this first step is really nice.
Maybe:
- Add a html5 video tag
- Check that video pointing by youtube/vimeo/facebook/... are encoded using a codec we can read on kiwix-desktop/kiwix-android/... to be sure that our failing tests are not about video encoding.

I don't know if this is the right place.

test-website/bad-encoding.html
benoit74: I mentioned the html meta but not the firefox thing, this is too specific and subject to change

test-website/chinese-encoding.html
benoit74: This is specific to one browser, is it really worth it to mention this?

benoit74:
> Add a html5 video tag
Done

> Check that video pointing by youtube/vimeo/facebook/... are encoded using a codec we can read on kiwix-desktop/kiwix-android/... to be sure that our failing tests are not about video encoding.

I have two issues with this:
- I don't know how to do this or even if this is feasible (but I could try / learn)
- I intentionally chose to do like most users would do on their websites: use directly the "embed" code generated by the source website (youtube/vimeo/facebook/...) so that we stay as close as possible to what we will encounter "in the wild"

For new HTML5 video I chose the VP8 webm video of Khan Academy (but specified the source MP4 as well, I don't know which encoder is used).

mgautierfr:
>  I intentionally chose to do like most users would do on their websites: use directly the "embed" code generated by the source website (youtube/vimeo/facebook/...) so that we stay as close as possible to what we will encounter "in the wild"

It is a rational counter-argument.

> For new HTML5 video I chose the VP8 webm video of Khan Academy (but specified the source MP4 as well, I don't know which encoder is used).

Maybe we can do the opposite, see what scrapperlib does with youtube/vimeo player and use the same encoding for our html5 video. If html5 video doesn't play, it means that it may not be a problem in js if youtube player doesn't work.

benoit74:
> Maybe we can do the opposite, see what scrapperlib does with youtube/vimeo player and use the same encoding for our html5 video. If html5 video doesn't play, it means that it may not be a problem in js if youtube player doesn't work.
But to be done in another PR anyway (if we do it)

I don't get it, VP8 webm is the format produced by scraperlib (this version of the video has been produced by scraperlib)

mgautierfr:
> I don't get it, VP8 webm is the format produced by scraperlib (this version of the video has been produced by scraperlib)

Haaa, everything is good then :)

benoit74:
@rgaudin do you wanna make a review or shall I merge this? (for now website update is manual, I will investigate how to automate this with the CI once this is merged, if you agree this is a good idea)

rgaudin:
Nope, go ahead

Issue #186, opened 2024-02-19
Failure of `MDN`: impossible to decode content


benoit74:
Last run of MDN recipe (https://farm.openzim.org/pipeline/961a1716-045f-4dc3-a5b3-dc8eb14315a8) failed with a problem while decoding a file.

```
Traceback (most recent call last):
  File "/usr/bin/zimit", line 8, in <module>
    sys.exit(zimit.zimit())
             ^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/zimit/zimit.py", line 611, in zimit
    run(sys.argv[1:])
  File "/app/zimit/lib/python3.11/site-packages/zimit/zimit.py", line 509, in run
    return warc2zim(warc2zim_args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/main.py", line 90, in main
    return converter.run()
           ^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/converter.py", line 264, in run
    self.add_items_for_warc_record(record)
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/converter.py", line 493, in add_items_for_warc_record
    payload_item = WARCPayloadItem(
                   ^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/items.py", line 38, in __init__
    (self.title, self.content) = Rewriter(path, record, known_urls).rewrite(
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/content_rewriting/generic.py", line 44, in rewrite
    return self.rewrite_html(head_template, css_insert)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/content_rewriting/generic.py", line 81, in rewrite_html
    to_string(self.content, self.encoding)
  File "/app/zimit/lib/python3.11/site-packages/warc2zim/utils.py", line 111, in to_string
    raise ValueError(f"Impossible to decode content {input_[:200]}")
[...]

PR #187, opened 2024-02-19
Better Exception, with item's path, when we cannot decode the content.

mgautierfr:
See suggestion in #186 about better error message.


benoit74 (approved):
LGTM, thank you

PR #192, opened 2024-02-28
Correctly rewrite json.

### LINKED ISSUE #160: Zimit2: Should we rewrite JSON?
See https://github.com/openzim/warc2zim/pull/158#discussion_r1467769011 for details

### LINKED ISSUE #181: [zimit2] Dynamically generated content derived from JSON not always showing in zimit2 version of MDN
Comparing the zimit2 run of MDN (https://dev.library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-02/) and last month's zimit classic run (https://library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-01), some of the dynamically generated information, specifically the browser compatibility tables at the end of each JS API article, are missing in the zimit2 version. Please see the screenshot below. Although the screenshot uses the PWA, I have verified, and you can check, the same issue on Kiwix Serve by comparing these two links and scrolling to the bottom of each page: [zimit1 showOPenFilePicker article](https://library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-01/A/developer.mozilla.org/en-US/docs/Web/API/Window/showOpenFilePicker) -> [zimit2 showOpenFilePicker article](https://dev.library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-02/developer.mozilla.org/en-US/docs/Web/API/Window/showOpenFilePicker).

image

Looking at console.log, it seems the error is caused by this `JSON.parse()` exception. Possibly a Wombat-generated URL that is not correctly interpreted?

image

mgautierfr:
Inline json was rewriten as javascript. Now we rewrite it as json.

This PR also rewrite "plan" json file with domain specific rules and rewrite "jsonp" (json callback) content.
This two fixes come from wabac rewriting but don't seems to be necessary (at least for this fix) and are not tested.

You can find a mdn archive converted to zim using this PR at https://tmp.kiwix.org/ci/test-warc/mgautier/mdn_json.zim

Ping @jaifroid

Fix #181
Fix #160

### DIFF
--- a/src/warc2zim/content_rewriting/ds.py
+++ b/src/warc2zim/content_rewriting/ds.py
@@ -8,10 +8,10 @@
 
 import json
 import re
+from collections.abc import Iterable
 from dataclasses import dataclass
 
 from warc2zim.content_rewriting.js import (
-    JsRewriter,
     TransformationAction,
     TransformationRule,
     add_prefix,
@@ -160,6 +160,8 @@ def rule_rewrite_vimeo_dash_manifest(m_object: re.Match, opts: dict | None) -> s
     return json.dumps(vimeo_manifest)
 
 
+# This set of rules tell which rules to apply depending on the url of the content
+# First rule to match stops the lookup, so rules have to be sorted accordingly.
 RULES = [
     SpecificRules(
         ["youtube.com", "youtube-nocookie.com"],
@@ -256,14 +258,15 @@ RULES = [
 ]
 
 
-def build_domain_specific_rewriter(path: str, url_rewriter):
+def get_ds_rules(path: str) -> Iterable[TransformationRule]:
     """
-    This build a JsRewriter with extra rules depending of the path.
+    This build a domain specifc Rewriter (given in rewriter_class), by passing a set of
+    extra rules to the rewriter if needed.
     """
 
     for rule in RULES:
-        for contain in rule.contains:
-            if contain in path:
-                return JsRewriter(url_rewriter, rule.rx_rules)
+        for pattern in rule.contains:
+            if pattern in path:
+                return rule.rx_rules
 
-    return JsRewriter(url_rewriter)
+    return []
--- a/src/warc2zim/content_rewriting/generic.py
+++ b/src/warc2zim/content_rewriting/generic.py
@@ -1,3 +1,4 @@
+from collections.abc import Callable
 from urllib.parse import urlsplit
 
 from jinja2.environment import Template
@@ -14,6 +15,25 @@ from warc2zim.utils import (
     get_record_url,
     to_string,
 )
+
+
+def no_title(
+    function: Callable[..., str | bytes]
+) -> Callable[..., tuple[str, str | bytes]]:
+    """Decorator for methods transforming content without extracting a title.
+
+    The generic rewriter return a title to use as item title but most
+    content don't have a title. Such rewriter just rewrite content and do not
+    extract a title, but the general API of returning a title must be fulfilled.
+
+    This decorator takes a rewriter method not returning a title and make it return
+    an empty title.
+    """
+
+    def rewrite(*args, **kwargs) -> tuple[str, str | bytes]:
+        return ("", function(*args, **kwargs))
+
+    return rewrite
 
 
 class Rewriter:
@@ -88,12 +108,11 @@ class Rewriter:
             self.content_str
         )
 
-    def rewrite_css(self):
-        return ("", CssRewriter(self.url_rewriter).rewrite(self.content))
+    @no_title
+    def rewrite_css(self) -> str | bytes:
+        return CssRewriter(self.url_rewriter).rewrite(self.content)
 
-    def rewrite_js(self, opts):
+    @no_title
+    def rewrite_js(self, opts: dict[str, Any]) -> str | bytes:
         rewriter = build_domain_specific_rewriter(self.path, self.url_rewriter)
-        return (
-            "",
-            rewriter.rewrite(self.content_str, opts),
-        )
+        return rewriter.rewrite(self.content.decode(), opts)
--- a/src/warc2zim/content_rewriting/generic.py
+++ b/src/warc2zim/content_rewriting/generic.py
@@ -1,12 +1,14 @@
 from collections.abc import Callable
+from typing import Any
 from urllib.parse import urlsplit
 
 from jinja2.environment import Template
 from warcio.recordloader import ArcWarcRecord
 
 from warc2zim.content_rewriting.css import CssRewriter
-from warc2zim.content_rewriting.ds import build_domain_specific_rewriter
+from warc2zim.content_rewriting.ds import get_ds_rules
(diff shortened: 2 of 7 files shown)


Jaifroid:
Thank you, @mgautierfr I've tested that archive in the PWA, and both of the test articles I mentioned are now fixed. Screenshot shows the article "Performance Web API". On the left is the previous zimit1 scrape which doesn't show the Browser Compatibility table, and on the right is the new test zimit2 ZIM which does. So, it seems to be fixed!

image

src/warc2zim/content_rewriting/generic.py
```diff
@@ -15,6 +20,29 @@
...
 
+# Parse JSONP. This match "anything" preceded by space or comments
+JSONP_REGEX = re.compile(
+    r"^(?:\s*(?:(?:\/\*[^*]*\*\/)|(?:\/\/[^\n]+[\n])))*\s*([\w.]+)\([{[]"
+)
+JSONP_CALLBACK_REGEX = re.compile(r"[?].*(?:callback|jsonp)=([^&]+)", re.I)
+
+
+def no_title(
+    function: Callable[..., str | bytes]
```
rgaudin, line 31: Needs a dosctring

src/warc2zim/content_rewriting/generic.py
```diff
@@ -15,6 +20,29 @@
...
+
+    return rewrite
+
+
+def extract_jsonp_callback(url: str):
+    callback = JSONP_CALLBACK_REGEX.match(url)
+    if not callback or callback[1] == "?":
+        return None
+
+    return callback[1]
```
rgaudin, line 44: Not fan of positional matching but those regexp are already too complex! 😅

src/warc2zim/content_rewriting/generic.py
```diff
@@ -88,12 +135,36 @@ def rewrite_html(self, head_template: Template, css_insert: str | None):
...
+        match = JSONP_REGEX.match(content)
+        if not match:
+            return content
+
+        callback = extract_jsonp_callback(self.orig_url_str)
+
+        if not callback:
+            return content
+
+        return callback + match.group(1)
```
rgaudin, line 160: Let's be consistent and either use `match[1]` here or use `m.group(1)` above

src/warc2zim/content_rewriting/generic.py
```diff
@@ -88,12 +135,36 @@ def rewrite_html(self, head_template: Template, css_insert: str | None):
...
+
+    @no_title
+    def rewrite_json(self):
+        content = self.rewrite_jsonp()[1]
```
rgaudin, line 164: Isn't it returning a string? I notice many methods not declaring a return type. We should.

rgaudin (changes requested):
Looks pretty good

src/warc2zim/content_rewriting/generic.py
mgautierfr: The method return a string, but it is decorated with `no_title` which makes it return a
tuple `("", rewritten_content)`.

benoit74: Then the type hint of the method is `str`, the decorator has its own type hint.

src/warc2zim/content_rewriting/ds.py
```diff
@@ -256,14 +255,15 @@ def filter_by_bitrate(array, max_bitrate, mime):
...
-def build_domain_specific_rewriter(path: str, url_rewriter):
+def get_ds_rules(path: str):
     """
-    This build a JsRewriter with extra rules depending of the path.
+    This build a domain specifc Rewriter (given in rewriter_class), by passing a set of
+    extra rules to the rewriter if needed.
     """
 
     for rule in RULES:
         for contain in rule.contains:
```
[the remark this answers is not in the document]
mgautierfr: Done

src/warc2zim/content_rewriting/generic.py
mgautierfr: Use `m.group(1)` everywhere.

src/warc2zim/content_rewriting/generic.py
mgautierfr: Yes, I have added the type hint. But we still need to do the indexation `[1]` as the decorated `rewrite_jsonp` returns a tuple.

mgautierfr:
Ready for a new review.

benoit74:
This PR will fix https://github.com/openzim/warc2zim/pull/160 as well or is there anything left?

benoit74 (approved):
LGTM, I will add the JSON and JSONP cases to the test website of https://github.com/openzim/warc2zim/pull/184

benoit74:
@rgaudin do you wanna review it again or shall we merge?

src/warc2zim/content_rewriting/ds.py
```diff
@@ -256,14 +255,15 @@ def filter_by_bitrate(array, max_bitrate, mime):
 ]
 
 
-def build_domain_specific_rewriter(path: str, url_rewriter):
+def get_ds_rules(path: str):
```
rgaudin, line 258: We have no return type, while this returns conditionnaly

src/warc2zim/content_rewriting/ds.py
```diff
@@ -256,14 +255,15 @@ def filter_by_bitrate(array, max_bitrate, mime):
...
+    extra rules to the rewriter if needed.
     """
 
     for rule in RULES:
-        for contain in rule.contains:
-            if contain in path:
-                return JsRewriter(url_rewriter, rule.rx_rules)
+        for pattern in rule.contains:
+            if pattern in path:
+                return rule.rx_rules
```
rgaudin, line 267: This excludes matching multiple rules. Is this on purpose? I can't think of any example of a single URL matching several rules but then maybe we should add a comment on top of `RULES` to indicate that rules must be sorted by higher precision first (although I understand those are simply copied from upstream here)

src/warc2zim/content_rewriting/ds.py
```diff
@@ -256,14 +255,15 @@ def filter_by_bitrate(array, max_bitrate, mime):
...
 
-    return JsRewriter(url_rewriter)
+    return None
```
rgaudin, line 269: Why not returning an empty list? Would make return type unconditional

rgaudin (changes requested):

src/warc2zim/content_rewriting/ds.py
mgautierfr: Return type added

src/warc2zim/content_rewriting/ds.py
mgautierfr: Empty list returned.
But the conditional on the call site is mainly for performance purpose, no need to run a RxRewriter if there is no rules.

rgaudin: Boolean test of empty list is false as well and as good in this case

rgaudin (approved):

PR #194, opened 2024-02-29
Do not unescape html entities during our rewriting.

### LINKED ISSUE #182: Unwanted redirection to 404 occurs in several articles in MDN (on dev.library.kiwix.org)
There is unwanted redirection of (correctly loaded) articles in the February MDN ZIM on dev.library.kiwix.org (also happens in PWA). A number of articles (which I've verified to be in the ZIM) get redirected nearly instantly to a custom Mozilla 404 page. If you're quick, you can just see a flash of the original content before it gets redirected.

Here's an example:

> Go to this page:
> 
> https://dev.library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-02/developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web
> 
> and click on "JavaScript Basics" in the left-hand menu. If you're quick, you'll see the page loads, but then very quickly redirects to a custom Mozilla 404 page.
> 
> This doesn't happen in the Zimit1 version of the same page here:
> 
> https://library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-01/A/developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web

NB I can see the content of such pages has been cached, and is all there and correct. The DevTools Cache API page even shows a graphical view of the requested page (see screenshot). The redirect definitely happens after DOMContentReady as a result of a dynamic Fetch request (possibly one of the many POST requests).

By comparison, in the January zimit1 version (on library.kiwix.org) there are several Header lookups, and also some attempt to do fuzzy matching in the Service Worker after the content has loaded, but no redirect occurs.

The exact cause of the redirect needs debugging.

image

_Originally posted by @Jaifroid in https://github.com/openzim/warc2zim/issues/178#issuecomment-1937914479_

mgautierfr:
Fix #182

The root cause was that we are decoding html entities (`&gt;`) to (`>`).
With such content `<code>&lt;script&gt;</code>` being rewritten as `<code><script></code>`.

Ping @jaifroid

mdn zim file build with this PR (and #192, even if this PR is directly based on warc2zim2) can be found at https://tmp.kiwix.org/ci/test-warc/mgautier/mdn_json_noescape.zim

### DIFF
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -59,7 +59,7 @@ class HtmlRewriter(HTMLParser):
         pre_head_insert: str,
         post_head_insert: str | None,
     ):
-        super().__init__()
+        super().__init__(convert_charrefs=False)
         self.url_rewriter = url_rewriter
         self.css_rewriter = CssRewriter(url_rewriter)
         self.title = None
@@ -125,6 +125,12 @@ class HtmlRewriter(HTMLParser):
             if data.strip():
                 data = JsRewriter(self.url_rewriter).rewrite(data)
         self.send(data)
+
+    def handle_entityref(self, name: str):
+        self.send(f"&{name};")
+
+    def handle_charref(self, name: str):
+        self.send(f"&#{name};")
 
     def handle_comment(self, data: str):
         self.send(f"<!--{data}-->")


rgaudin (approved):
LGTM

benoit74 (approved):
LGTM as well

PR #196, opened 2024-02-29
Main page might be a redirect | ZIM illustration/favicon is not retrieved properly

### LINKED ISSUE #193: Handle case where main url is a redirect in warc2zim instead of zimit
Previously, the fact that the url passed leads to a redirect page was handled in zimit, i.e zimit checked first if a redirect was happening before passing the result of the redirection to warc2zim.

This caused issues reported in https://github.com/openzim/zimit/issues/256 since the check for redirection was done in Python.

Since browsertrix crawler 1.0.0-beta5 (released today), browsertrix crawler handles the redirection on its side, creating an additional seed the URL is redirected. This present the big advantage that scope issues do not apply anymore due to the additional seed which will be considered for prefix scope matching for instance.

This means that we would now probably like to get rid of `check_url` in zimit (keep only the cleanup of 443 and 80 default ports) but it means that warc2zim has to be slightly adapted: it is possible that the first WARC record is a simple redirect, and the main url, title, ... are hence not the ones of the first record, but the one of the second record.

### LINKED ISSUE #195: Favicon URL is not retrieved when not passed at CLI and scheme is not specified in HTML document
When no favicon URL is passed as a CLI input and we try to "guess" it from html main page headers and the scheme is not specified, the resulting favicon URL is incorrect.

Not working:
```
<link rel="icon" href="//dha92jo6cen2v.cloudfront.net/styles/mefi/favicon030514b.ico" type="image/ico">
```

```
<link rel="icon" href="/styles/mefi/favicon030514b.ico" type="image/ico">
```

```
<link rel="icon" href="./styles/mefi/favicon030514b.ico" type="image/ico">
```

The cause is that the scraper tries to build the favicon URL with an `urljoin` on `self.main_url` and the value extracted from the link above, but `self.main_url` is normalized and hence does not contain the HTTP scheme anymore.

benoit74:
## Main page might be a redirect

Fix #193 

The main page of the WARC might be a redirect, especially since browsertrix crawler 1.0.0-beta5 is handling them properly (https://github.com/webrecorder/browsertrix-crawler/pull/476) and we plan to move to this (see https://github.com/openzim/zimit/issues/256).

The logic is hence slightly changed to process this situation, i.e we check the status code of the main record, and if it is a 301 or 302 we process the Location to find the warc record matching the redirect. This is done iteratively, so many redirects are probably handled (this has not been tested, I do not have an example website available).

## ~~Favicons are missing the HTTP scheme~~

~~Fix #195 ~~

~~We simply store the main URL HTTP scheme, and update it in case of redirection.~~

**UPDATE:**

## Favicon (ZIM illustration) is not retrieved properly

Fix #195 

Changes:

- make a distinction between the case where the favicon URL is passed as (CLI) argument and the case where we guess the favicon path from main page icon
- download illustration only if favicon URL is passed
  - in other words, do not download favicon path if not present inside the WARC (no scenario where it would make sense)


benoit74:
Ideas on how to test this properly are welcomed ? Should I just add the test WARC from https://metafilter.com/ I used in the codebase - I could limit it to just 5 records ?

src/warc2zim/converter.py
```diff
@@ -319,6 +321,22 @@ def gather_information_from_warc(self):
 
             # if we get here, found record for the main page
 
+            # if main page is a redirect, update the main url accordingly
+            if record.http_headers:
+                status_code = int(record.http_headers.get_statuscode())
+                if status_code in [
```
rgaudin, line 327: I think we should handle the case of multiple redirects

src/warc2zim/converter.py
```diff
@@ -319,6 +321,22 @@ def gather_information_from_warc(self):
...
+                    HTTPStatus.MOVED_PERMANENTLY,
+                    HTTPStatus.FOUND,
+                ]:
+                    redirected_url = record.http_headers.get_header("Location")
+                    self.main_url_scheme = urlsplit(redirected_url).scheme
+                    self.main_url = normalize(redirected_url)
```
rgaudin, line 333: If it doesn't include the scheme, this is misleading. I don't think we need this. Is this only to handle a special case for favicon? Then it shouldn't be there

src/warc2zim/converter.py
```diff
@@ -362,13 +380,15 @@ def find_icon_and_language(self, content):
                 )
             ):
                 self.favicon_url = urljoin(
-                    self.main_url,
+                    f"{self.main_url_scheme}://{self.main_url}",
```
rgaudin, line 383: You should be able to detect if the href is a scheme-less (same scheme) URL instead of a path and just recompose reusing the main_url's scheme. That would be more reasonnable

rgaudin (changes requested):

src/warc2zim/converter.py
benoit74: Nevermind, I've found a pretty good test URL : `https://www.kiwix.org/`, which redirects to `https://www.kiwix.org/en/` and then `https://kiwix.org/en/`

src/warc2zim/converter.py
benoit74: `main_url` is normalized since the beginning of zimit2 I think, and normalization removes the scheme.

`main_url_scheme` is for now only used for favicon, but we might need it for other things I think.

If it shouldn't be there, then where would you like to have it? If we are redirected from http to https on the main page, we probably need to know it, isn't it?

src/warc2zim/converter.py
rgaudin: 👍

src/warc2zim/converter.py
rgaudin: OK, it's a naming issue then. We can't keep a `redirected_url` that's a regular URL and a `main_url` that's a scheme-less normalized URL next to each other and expect that people know those are different.

I think it would make sense to use `normalized_main_url` for the normalized version but if it's all over the place, we could do the other way around.
keeping normalized and scheme separately and having a property returning the main_url is fine as well.

Can you look into what would be the most practical?

src/warc2zim/converter.py
benoit74: In fact browsertrix crawler is already "fully" resolving the redirect, so there will be only one redirect in all situations:

```
{"timestamp":"2024-03-01T09:35:45.493Z","logLevel":"info","context":"worker","message":"Starting page","details":{"workerid":0,"page":"http://www.kiwix.org/"}}
{"timestamp":"2024-03-01T09:35:45.494Z","logLevel":"info","context":"crawlStatus","message":"Crawl statistics","details":{"crawled":0,"total":1,"pending":1,"failed":0,"limit":{"max":10,"hit":false},"pendingPages":["{\"seedId\":0,\"started\":\"2024-03-01T09:35:45.412Z\",\"extraHops\":0,\"url\":\"http://www.kiwix.org/\",\"added\":\"2024-03-01T09:35:45.031Z\",\"depth\":0}"]}}
{"timestamp":"2024-03-01T09:35:46.485Z","logLevel":"info","context":"general","message":"Awaiting page load","details":{"page":"http://www.kiwix.org/","workerid":0}}
{"timestamp":"2024-03-01T09:35:47.645Z","logLevel":"info","context":"general","message":"Seed page redirected, adding redirected seed","details":{"origUrl":"http://www.kiwix.org/","newUrl":"https://kiwix.org/en/","seedId":1}}
```

rgaudin: Good to know. Should we document it in the code?

src/warc2zim/converter.py
benoit74: `main_url` is always scheme-less in current design

src/warc2zim/converter.py
benoit74: I've moved another way, computing the scheme only when needed, i.e. in `gather_information_from_warc` ; we do not need this information elsewhere in fact

src/warc2zim/converter.py
benoit74: I was misled by Browsertrix logs, there are many redirects in the WARC (and it is handled mostly nicely, just had to fix a small bug due to the fact that "Location" header might be a relative path only, not a full URL)
```
--------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------
[warc2zim::2024-03-01 11:21:07,708] WARNING:HTTP 301 occurred on main page; replacing http://www.kiwix.org/ with 'https://www.kiwix.org/'
[warc2zim::2024-03-01 11:21:07,709] WARNING:HTTP 301 occurred on main page; replacing https://www.kiwix.org/ with 'https://www.kiwix.org/en/'
[warc2zim::2024-03-01 11:21:07,709] WARNING:HTTP 301 occurred on main page; replacing https://www.kiwix.org/en/ with 'https://kiwix.org/en/'
```

benoit74:
I've added more commits to:
- simplify logic around `main_url_scheme` / `scheme`
- add a test WARC and test case for `kiwix.org` which performs multiple redirects, including a single redirect from HTTP to HTTPS (and there are one WARC record per redirect)
- fix handling of redirect of main page when the `Location` header is only a relative path
- fix the test logic which was not really checking that favicon is the appropriate one, just that "there is a favicon"
- fix the issue around the fact that favicon was always downloaded even if present inside the WARC (luckily the one from reseau-canope WARC is now gone, so it raised an issue when checking the favicon)

Please review again

benoit74:
I've rename `main_url` to `main_path` for even more clarity, sorry for the commit after request for review but it's a small (but important) semantic change.

src/warc2zim/converter.py
```diff
@@ -228,7 +228,7 @@ def run(self):
 
         self.creator = Creator(
             self.full_filename,
-            main_path=self.main_url,
```
benoit74, line 230: This is where I got the "proof" it should be named `main_path`

rgaudin (approved):
👍  thank you

mgautierfr:
I'm a bit disturbed by this PR. Not the PR itself, but I think the error is elsewhere.

> main_url_scheme is for now only used for favicon, but we might need it for other things I think.

And this sentence make me think you don't have the root cause neither :)
Why we need the scheme for the favicon ? This is the same thing than any other resource, we can use relative link and we don't need scheme for them. Favicon is not different. Or we use favicon elsewhere ?

I think we need the scheme for this line : https://github.com/openzim/warc2zim/blob/main/src/warc2zim/converter.py#L442, ie: when we download the favicon if we cannot found in the warc.
But I feel uncomfortable with this feature. I tend to think that a tool like warc2zim should be able to work offline, no download must be done. If a warc doesn't contains a favicon it is a scrapper problem. warc2zim should not try to be smart here.

rgaudin:
> If a warc doesn't contains a favicon it is a scrapper problem. warc2zim should not try to be smart here.

I agree. I'm not sure why we had the feature in the first place but there's not real argument for it. We still need to parse to retrieve URl and from it the WARC record then extract it. If any of this fails, we have the PNG fallback.

benoit74:
> Why we need the scheme for the favicon ? This is the same thing than any other resource, we can use relative link and we don't need scheme for them. Favicon is not different. Or we use favicon elsewhere ?

We need this for two reasons:
- the one you've found about the download fallback (we might remove this, please open an issue to make it more visible)
- but also because we use `urljoin` to join the `main_path` with the `favicon_url`, and this fails if scheme is not provided in `main_path` ; I wonder how we currently compute the "real" URL from a `src` or `href` on any given page, since this is the same logic ; I will have a look

> If a warc doesn't contains a favicon it is a scrapper problem. warc2zim should not try to be smart here.

I don't agree (no drama), let's discuss it here: https://github.com/openzim/warc2zim/issues/202

benoit74:
In fact, in other places of the code base we use the `ArticleUrlRewriter` to compute "real" URLs. And this rewriter is based on the `orig_url_str`, which is the original WARC record URL and hence contains the scheme, and call `urljoin` on `origin_url_str` and the `src` / `href` value (and then normalize, and handle data and blob cases, ...).

Here for favicon we take a shortcut by calling `urljoin` directly because we are not at all in the WARC records processing logic, we are at the very beginning of the scraper. I feel like this shortcut is not a big deal, especially if (as I suggest in https://github.com/openzim/warc2zim/issues/202) we decide to not get rid of the "wise" online download fallback.

mgautierfr:
> but also because we use urljoin to join the main_path with the favicon_url, and this fails if scheme is not provided in main_path ; I wonder how we currently compute the "real" URL from a src or href on any given page, since this is the same logic ; I will have a look

The real question is why we do this urljoin ?

`self.favicon_url` is used in two places:
- https://github.com/openzim/warc2zim/blob/main/src/warc2zim/converter.py#L423 to compare the (found) favicon url with the record's url. We could ensure that both have correct scheme or simply normalize on them first. I'm not totally sure on this, but as we work all the time on normalized url, I think is would be better to work on normalized url.
- https://github.com/openzim/warc2zim/blob/main/src/warc2zim/converter.py#L442 to download the favicon. Here we have the discussion #202 open, but we could simply download the illustration given in the CLI and set `self.illustration`.

---

I don't think we will merge this PR at it is before #202 is resolved. Maybe you could split this PR in two and so we could discuss them individually.

benoit74:
I think we have a kind of consensus on https://github.com/openzim/warc2zim/issues/202, we shouldn't download what is most obviously present in the WARC.

For favicon, we do the urljoin because we have to transform any relative URL (`//myotherhost.com/myimage.png`, `../myimage.png`, ...) into the full URL, like it is done for every `src` / `href` in `ArticleUrlRewriter`

I made a force push (sorry) with following changes (compared to your last review):
- rewrite everything into only four commits to clean everything up / separate concerns
- re-implemented the favicon logic to:
  - compute the proper favicon path based on same logic as `ArticleUrlRewriter` (but without the `from_normalized` operation which make all paths relative to current article)
  - make a distinction between `favicon_url` (URL passed as argument) and new `favicon_path` (retrieved from main page icon, must be present in WARC, no download please)
- I do not need the HTTP scheme anymore (in fact, it is kinda trick since it is present in `get_record_url`, but this is exactly what `ArticleUrlRewriter` is doing

src/warc2zim/converter.py
```diff
@@ -398,42 +427,47 @@ def find_icon_and_language(self, content):
...
+                    if (
+                        record.http_headers
+                        and record.http_headers.get_statuscode() != "200"
+                    ):  # pragma: no cover
+                        logger.warning("WARC record for favicon is unusable")
+                        break
+                    self.illustration = get_record_content(record)
+                    break
+
+            # download favicon_url (might be custom URL, not present in WARC records)
```
rgaudin, line 452: I understood this should have been removed?

src/warc2zim/converter.py
```diff
@@ -421,42 +427,47 @@ def find_icon_and_language(self, content):
...
+                if path == self.favicon_path or url == self.favicon_url:
+                    logger.debug("Found WARC record for favicon")
+                    if (
+                        record.http_headers
+                        and record.http_headers.get_statuscode() != "200"
+                    ):  # pragma: no cover
+                        logger.warning("WARC record for favicon is unusable")
+                        break
+                    self.illustration = get_record_content(record)
+                    break
```
mgautierfr, line 450: Should we have a `return` or something like that to not download the favicon just after ?

src/warc2zim/converter.py
benoit74: look at next line, `favicon_url` is set only by input (CLI) argument

mgautierfr (changes requested):

src/warc2zim/converter.py
benoit74: Good point, I fixed it slightly differently since I do want to `del` the variables.

rgaudin (approved):
👍

rgaudin:
PR should be renamed though

benoit74:
PR title and first comment updated to better reflect the final content of this PR

benoit74:
@mgautierfr anymore remark / change request?

mgautierfr (approved):
We are good.
Please do the rebase-fixup and you can directly merge the PR.

PR #198, opened 2024-02-29
Ignore `resource` WARC records for now

### LINKED ISSUE #197: Do not include `resource` WARC records inside the ZIM
https://github.com/webrecorder/browsertrix-crawler/issues/457 and https://github.com/webrecorder/browsertrix-crawler/pull/458 have introduced more systematic use of `resource` WARC records.

As of 1.0.0-beta5, `resource` WARC records are used for:
- thumbnails
- screenshots
- text extracts
- page infos (mentioned above)

All these resources while potentially useful for warc2zim logic (especially the page infos, we do not use thumbnails, screenshots and text extracts for now) should not be directly embedded inside the ZIM, at least not as a response record is.

benoit74:
Fix #197 

`add_items_for_warc_record` is already called only for records of type `resource`, `response` and `revisit` (i.e. `request` records are ignored) due to the filtering logic of `iter_warc_records`.

The logic to add a real item to the ZIM was not called if record type is `revisit`, i.e. it was called only for `resource` and `response`. 

As explained in #197, `resource` WARC records must not be added as-is to the ZIM, they are not regular web responses (as their WARC type implies).

The logic is hence inverted to also be future-prood:
- apply the default logic to add an item to the ZIM only if record is a `response`
- apply the logic to compute revisits only if record is a `revisit`

### DIFF
--- a/tests/test_warc_to_zim.py
+++ b/tests/test_warc_to_zim.py
@@ -159,7 +159,7 @@ class TestWarc2Zim:
                 # We should check with the content of the targeted record...
                 # But difficult to test as we don't have it
                 assert payload
-            else:
+            elif record.rec_type == "response":
                 # We must have a payload
                 assert payload
                 payload_content = payload.content.tobytes()
@@ -168,6 +168,9 @@ class TestWarc2Zim:
                 # have exact match
                 if payload.mimetype.startswith("text/html"):
                     assert head_insert in payload_content
+            elif record.rec_type == "resource":
+                # we do not want to embed resources "as-is"
+                assert not payload
 
             warc_urls.add(url)
 
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -485,7 +485,7 @@ class Converter:
                 logger.debug(f"Skipping url {url}, outside included domains")
                 return
 
-        if record.rec_type != "revisit":
+        if record.rec_type == "response":
             if self.is_self_redirect(record, url):
                 logger.debug("Skipping self-redirect: " + url)
                 return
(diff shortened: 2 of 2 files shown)


rgaudin (approved):
LGTM

tests/test_warc_to_zim.py
```diff
@@ -168,6 +168,9 @@ def verify_warc_and_zim(self, warcfile, zimfile, verify_scraper_suffix):
                 # have exact match
                 if payload.mimetype.startswith("text/html"):
                     assert head_insert in payload_content
+            elif record.rec_type == "resource":
+                # we do not want to embed resources "as-is"
+                assert not payload
```
mgautierfr, line 173: Should we really fail if a resource has a payload ?
It open a way to have a change in a part we don't use breaking our code.

mgautierfr (changes requested):

tests/test_warc_to_zim.py
rgaudin: This is a test 👀

benoit74: And the test says: "we do not produce a ZIM item for WARC record of type `resource`", which must be true for now ; might change in the future if we decide to use `resource` records, but then we will fix the test intentionally because we know which kind of `resource` must produce which kind of `payload` ; and I even doubt we will embed them at the same path than the one used in WARC

mgautierfr: > This is a test 👀

🤦🏽 My bad. We are good then.

mgautierfr (approved):

PR #201, opened 2024-03-01
Upgrade to Python 3.12, Python dependencies and add hatch-openzim plugin

benoit74:
Maintenance work to move on ^^


tasks.py
```diff
@@ -92,7 +92,7 @@ def fix_black(ctx: Context, args: str = "."):
 def fix_ruff(ctx: Context, args: str = "."):
     """fix all ruff rules"""
     args = args or "."  # needed for hatch script
-    ctx.run(f"ruff --fix {args}", pty=use_pty)
+    ctx.run(f"ruff check --fix {args}", pty=use_pty)
```
rgaudin, line 95: Ah! https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#cli

rgaudin (approved):
LGTM

tasks.py
benoit74: Yep ... so minor I did not took time to mention it, but obviously I should have to spare your time ^^

PR #204, opened 2024-03-04
Use Lookahead expression in fuzzy rules instead of Lookbehind

### LINKED ISSUE #174: Incompatible regexp in fuzzy substitutions in `wombat_setup.js` blocks video in iOS Safari (at least)
As part of #144, I tested the "orca" video in 100rabbits (26th Jan version) on iOS Safari (using the PWA, via Browser Stack live testing). There is a blocker on this platform due to an incompatible regular expression. It seems the "googlevideo" regex (see highlighted line in screenshot) uses a syntax that is not supported in Safari, and stalls with the error "Invalid group specifier name". I didn't test Firefox or Chrome on iOS, but all browsers currently have to use the Safari engine on iOS. This might or might not affect the web view in the official app, but preliminary reports suggest it's not working there either.

I think this issue somehow is tangentially related to #171, because Babel can take care of ensuring cross-browser compatible syntax is used or polyfilled if necessary.

image

Ping @mgautierfr

mgautierfr:
Lookbehind (https://caniuse.com/?search=Lookbehind) is not fully supported on "old" browsers.
But lookahead (https://caniuse.com/?search=Lookahead) is supported "everywhere" as it is part of original js regular expression.
    
The tricky part in this change is that we must match the `?` both as:
- Just after `videoplayback` (always)
- Potentially just before `id=...`.

Fix #174 
Ping @jaifroid

### DIFF
--- a/src/warc2zim/url_rewriting.py
+++ b/src/warc2zim/url_rewriting.py
@@ -63,8 +63,8 @@ logger = logging.getLogger("warc2zim.url_rewriting")
 
 FUZZY_RULES = [
     {
-        "pattern": r".*googlevideo.com/(videoplayback\?).*((?<=[?&])id=[^&]+).*",
-        "replace": r"youtube.fuzzy.replayweb.page/\1\2",
+        "pattern": r".*googlevideo.com/(videoplayback(?=\?)).*[?&](id=[^&]+).*",
+        "replace": r"youtube.fuzzy.replayweb.page/\1?\2",
     },
     {
         "pattern": r"(?:www\.)?youtube(?:-nocookie)?\.com/(get_video_info\?).*(video_id"
--- a/tests/test_fuzzy_rules.py
+++ b/tests/test_fuzzy_rules.py
@@ -1,0 +1,41 @@
+import re
+
+import pytest
+
+from warc2zim.url_rewriting import FUZZY_RULES
+
+from .utils import ContentForTests
+
+
+@pytest.fixture(
+    params=[
+        ContentForTests(
+            "foobargooglevideo.com/videoplayback?id=1576&key=value",
+            "youtube.fuzzy.replayweb.page/videoplayback?id=1576",
+        ),
+        ContentForTests(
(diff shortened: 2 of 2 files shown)


src/warc2zim/url_rewriting.py
```diff
@@ -63,8 +63,8 @@
 
 FUZZY_RULES = [
     {
-        "pattern": r".*googlevideo.com/(videoplayback\?).*((?<=[?&])id=[^&]+).*",
-        "replace": r"youtube.fuzzy.replayweb.page/\1\2",
+        "pattern": r".*googlevideo.com/(videoplayback(?=\?)).*[?&](id=[^&]+).*",
```
Jaifroid, line 66: Maybe I've misunderstood, but how is this different from just doing:

`(videoplayback\?).*[?&]` (etc.)?

You don't need a lookahead, ISTM, if you simply want to match a question mark that will always be there. The `.*` after the lookahead will consume the matching `?` anyway. Only reason to use a lookahead is when you don't want to consume any characters but you want to know that there is something ahead.

Jaifroid: Or, if you don't want to match the question mark in the first parentheses, which looking at it I presume you don't, then put `\?` outside the first parens.

mgautierfr: > The `.*` after the lookahead will consume the matching `?`

It will consume **until** the `?` but it will not consume it.

And we don't want to consume it (first) to have the `[?&]` consume it.
With `(videoplayback\?)` we consume the `?` and so `[?&]`  will never match on `googlevideo.com/videoplayback?id=1234`

Jaifroid: Original pattern consumed the question mark inside the first parentheses...

Jaifroid: OK, I've understood it now! Thanks for the explanations!

Jaifroid (approved):
I checked the test cases in RegexBuddy, and also the non-matching cases. Looks good!

benoit74 (approved):
LGTM

PR #208, opened 2024-03-07
Add wombat.js back

### LINKED ISSUE #207: warc2zim2: warc2zim package does not contain the wombat.js anymore
The package does not contains the wombat.js anymore, which is quite embarrassing.

"Problem" is that hatch respect .gitignore: https://hatch.pypa.io/latest/config/build/#vcs

Issue has been introduced by https://github.com/openzim/warc2zim/commit/7dbc9d79a8047d62d706511dbac6ee7fded1aa09

I will fix it right now, most probably by adding the `statics` directory as artifact (it is still useful that hatch respects .gitignore for many other reasons)

benoit74:
Fix #207 

Changes:
- include all `statics` files as artifacts for the wheel

Nota:
- the sdist still doesn't contain the `wombat.js` which is great

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,9 @@ exclude = [
 
 [tool.hatch.build.targets.wheel]
 packages = ["src/warc2zim"]
+artifacts = [
+  "src/warc2zim/statics/**",
+]
 
 [tool.hatch.envs.default]
 features = ["dev"]


rgaudin (approved):
This looks like the proper way to do it then. Can we consider it the final solution and the way to replicate in other scrapers?

PR #213, opened 2024-03-14
Url encoding tests - more

benoit74:
Even more tests for URL encoding - this time with querystring which leads to different images.

I had to create a "webserver code" (here PHP script on an external domain) to dynamically process these querystring and provide a different image.

I begin to consider we have reached the limit of the "quick-and-dirty" upload to a simple webserver, I don't like the idea that our tests depends on an external server.

I would suggest that:
- we build a Docker image for the test-website
- we run this image in our k8s, and serve it with two domains:
  - `test-website.dev.kiwix.org` => the main test website
  - `tést-wébsite.dev.openzim.org` => a "fake" external website with utf-8 chars in domain name

I don't yet know if using PHP is the proper solution or if I should convert this to Python.

WDYT?


rgaudin (approved):
LGTM

benoit74:
@rgaudin WDYT about my suggestion to build a custom Docker image and host it in k8s?

rgaudin:
Probably the way to go

benoit74:
Oups, I forgot your suggestion. Sorry. Repo is not correctly configured, I should has been forced to force the merge since discussions are not all resolved. I will fix it directly inside warc2zim2 branch.

Issue #206, opened 2024-03-06
Zimit2: Fix URL encoding of ZIM items


kelson42:
Original page:
https://thalesdocs.com/gphsm/luna/7/docs/network/Content/Home_Luna.htm

ZIM page:
https://dev.library.kiwix.org/viewer#thalesdoc_en_all_2024-03/thalesdocs.com/gphsm/luna/7/docs/network/Content/Home_Luna.htm

In the ZIM, all the PDF are missing, although nothing special about them is to remark from my POV: links are straight links to PDF files in the same mirrored domain.

Remark: PDF filenames have spaces!

Jaifroid:
Just curious as to whether this is a situation we need to handle. Clearly there's a web design issue on that site in that they forgot to URI-encode the URLs. Browsers are lax enough to work around that. The failure point could be anywhere: in the crawler, in the Archiving software (Replay), or in warc2zim, or even in the reader (Kiwix Serve or other reader)... We should probably double-check the problem isn't in our readers.

rgaudin:
ZIM has plenty of PDF:

```
❯ zimls thalesdoc_en_all_2024-03.zim |grep -E "\.pdf$" |wc -l
     719
```

`Counter` says `application/pdf=616;`

rgaudin:
Entry `thalesdocs.com/gphsm/luna/7/docs/network/Content/PDF_Network/Product%20Overview.pdf` is in the ZIM but requesting it seems impossible due to the decoding issue

```sh
❯ curl -I 'https://dev.library.kiwix.org/content/thalesdoc_en_all_2024-03/thalesdocs.com/gphsm/luna/7/docs/network/Content/PDF_Network/Product%20Overview.pdf'
HTTP/2 404
```

Jaifroid:
Ah, so the issue may be reader-side. Oh no, yet another 2GB file to download in order to test properly.....

Jaifroid:
I confirm PDFs are in ZIM, but they can't be accessed in the PWA or Browser Extension either. Failure seems to be because the backend URI-decodes the URL before requesting it from the ZIM, but unfortunately, as you see from screenshot the PDF URLs are actually store URI-encoded in the ZIM.

I think if these URLs were stored with spaces -- the ZIM spec does say that all URLs should be stored unencoded, but not sure where we stand on that wrt Zimit2 --- then I believe the readers would work, though it needs testing.

@mgautierfr?

[...]

PR #215, opened 2024-03-14
More fuzzy tests

benoit74:
Add way more tests around fuzzy rules.

This is preparatory work for the new normalisation of URLs (#206)


rgaudin (approved):
LGTM ; much needed but I don't see how this relates to #206 which anyway doesn't clearly express an issue and its way forward.

PR #224, opened 2024-04-04
Dockerized test-website

benoit74:
Changes:
- Build and publish a docker image for the test website
  - New segregated and custom Github workflow, triggered on `test-website` tag for convenience
  - This new workflow will publish automatically on k8s once Docker image has been integrated over there 
- Some adaptations needed for docker operations
  - source domain names and other stuff from environment variable and patch files at container startup
  - moved most test-website files to subfolder for clarity


.github/workflows/TestWebsite.yaml
```diff
@@ -0,0 +1,32 @@
...
+          restrict-to: openzim/warc2zim
+          registries: ghcr.io
+          credentials:
+            GHCRIO_USERNAME=${{ secrets.GHCR_USERNAME }}
+            GHCRIO_TOKEN=${{ secrets.GHCR_TOKEN }}
+          repo_description: auto
+          repo_overview: auto
+          platforms: |
+            linux/amd64
+            linux/amd64
```
rgaudin, line 30: ?

.github/workflows/TestWebsite.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+    # TODO: update test website automatically
```
rgaudin, line 32: ?

test-website/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+RUN apk add --no-cache bash
+
+COPY entrypoint.sh /root/entrypoint.sh
+
+ENTRYPOINT [ "/root/entrypoint.sh" ]
+
+CMD [ "caddy", "run", "--config", "/etc/caddy/Caddyfile", "--adapter", "caddyfile"]
+
+# # #---
```
rgaudin, line 16: Does this have any value?

test-website/README.md
```diff
@@ -13,3 +13,28 @@ It contains all kind of situations we currently cope with:
...
+Build the docker image
+
+```
+docker build -t test-website .
+```
+
+Start the test website with appropriate environment variables.
+
+```
+docker run -p 8888:80 --rm --name test-website -e SITE_ADDRESS="test-website.local.oviles.info:80, xn--jdacents-v0aqb.local.oviles.info:80" -e STANDARD_HOSTNAME="http:\/\/test-website.local.oviles.info:8888" -e ACCENTUATED_HOSTNAME="http:\/\/jédéacçents.local.oviles.info:8888" test-website
```
rgaudin, line 29: Port doesn't respect our convention 😉

test-website/README.md
```diff
@@ -13,3 +13,28 @@ It contains all kind of situations we currently cope with:
...
+
+In the example above, the trick is that we have the following DNS records in place : `local.oviles.info A 127.0.0.1` and `*.local.oviles.info CNAME local.oviles.info`, meaning that any request to local.oviles.info or one of its subdomain will resolve to localhost IP 127.0.0.1 ; we use local ports 8080 for HTTP and 8443 for HTTPS.
+
+You can then open https://test-website.local.oviles.info:8888 in your favorite browser and run manual tests on this website (which uses the other one as sub-site for few resources on special domains with special characters).
+
+## Environments variables needed in Docker image
+
+- SITE_ADDRESS: the site address for Caddy operation ; should contain a non-accentuated and an accentuated hostnames for proper testing
+- STANDARD_HOSTNAME: the full hostname to the non-accentuated / standard hostname as seen from the browser (will be used in HTML/JS/... files) ; will be used by `sed` as replacement, so few chars (e.g. `/`) must be escaped
+- ACCENTUATED_HOSTNAME: the full hostname to the accentuated hostname as seen from the browser (will be used in HTML/JS/... files) ; will be used by `sed` as replacement, so few chars (e.g. `/`) must be escaped
```
rgaudin, line 40: What's the difference with previous one?

rgaudin (changes requested):

.github/workflows/TestWebsite.yaml
benoit74: Copy/paste, it's the same on all other workflows 🤦🏻
Who knows where it came from ... 🤷🏻‍♂️
Since we do not need another platform obviously, I will remove it everywhere for now.

.github/workflows/TestWebsite.yaml
benoit74: This is intentional, to mention that we will soon deploy automatically very soon (matter of hours or days) like we do on the zimfarm, nautilus web ui, ...

test-website/Dockerfile
benoit74: Nope

.github/workflows/TestWebsite.yaml
rgaudin: Ah! I don't remember seeing it before

test-website/README.md
benoit74: Which convention? The one which is not yet formally approved? 🤣

test-website/README.md
benoit74: I've updated the documentation with a clearer table (I hope) and sample values.

Btw, here I used a "personal" domain name, do you mind if we add required records to `openzim.org` (e.g. `local.openzim.org` and `*.local.openzim.org` so that it looks less suspicious / does not depends on something I'm the only one to manage)

rgaudin: Thank you it's a lot clearer now.

- Probably wise to refer to them as `netloc` and not `domain` as it's closer to what this is.
- Given there's no need to resolve, I don't think using openzim domain (or yours) makes sense. It creates a deceptive expectation. There are [reserved names](https://en.wikipedia.org/wiki/Special-use_domain_name) for this purpose. I don't think that having localhost-targetting entries in our zone is wise ; at least until it's code-driven

rgaudin (approved):
LGTM ; see unresolved conversation and conflict before merging obviously

test-website/README.md
benoit74: > Probably wise to refer to them as netloc and not domain as it's closer to what this is.
Agreed 

> Given there's no need to resolve ...
There is a need to resolve for local tests, this is why I mention they resolve to 127.0.0.1 and why I suggest these special values for local tests

rgaudin: it's a developer-only need (you must run it for it to work) and you've covered it in the README. This is superfluous to have it in the zone. We already have entries we don't remember what for so as I said, until we are code-driven (ie. with comments), I don't think it's worth it.

benoit74: OK, I'm fine with it.

PR #225, opened 2024-04-05
Javascript tests in test website

benoit74:
Changes:
- finish the renaming from hostname to netloc (only environment variables where renamed, not placeholders)
- add tests for Javascript (classic and module)


benoit74:
As agreed live, these small changes on the test website are not reviewed

Issue #219, opened 2024-03-18
Zimit2: Is there an HTML parser issue with some special characters?


benoit74:
Given following HTML code:

```
<img width="200" src="image.png?param1=value1&param2=value2" />
```

Current HTML parser (in warc2zim2 branch) decode the `src` tag as `image.png?param1=value1¶m2=value2`, i.e. it consider that `&para` is an entity reference even if it has not trailing `;` and decodes it as `¶`.

I've found https://github.com/python/cpython/issues/69426 which is more or less related, but did not saw much progress since years. One person on this issue states that the URL with `&para` is an invalid one, but I'm not convinced at all, at least W3C validator (e.g. https://validator.w3.org/nu/#textarea) is happy with following HTML document:

```
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="utf-8">
    <title>Characters encoding in a ZIM</title>
</head>

<body>
<img width="200" alt="" src="image.png?param1=value1&param2=value2">
</body>
</html>
```

This issue still has to be further analyzed, but I wonder if we should switch to another HTML parser. This could even help us benefit from more advanced parsing functionality which could help the scraper cope with poorly written HTML document (some parsers are better than others at recovering from bad code when building the tree).

benoit74:
Edit: updated first comment URL, I simplified only one of the two URLs ...

Jaifroid:
Briefly (I'm travelling) it looks like an over-enthusiastic parser, as it shouldn't be interpreting HTML entities in URLs, let alone in querystrings. (There is the possibility that our querystring obfuscation is somehow also involved.) Also, entities without `;` are surely not entities?!

benoit74:
> There is the possibility that our querystring obfuscation is somehow also involved.

Bug happens with just the HTML code I pushed in first commit with a bare html parser, no link with any manipulation done in warc2zim.

> Also, entities without ; are surely not entities?!

This is a debatable point of view according to Python folks, see upstream issue

benoit74:
We need anyway a better documentation of HTML rewriter at `src/warc2zim/content_rewriting/html.py` ; all functions are missing docstrings

benoit74:
## The specifications

In HTML, we are supposed to inherit from XML constraints saying that `&`, `<` and `>` in attributes values and data must be escaped. E.g. `&` must be written `&amp;`. At least this is what https://stackoverflow.com/a/7382028 suggests, but it lacks references to HTML standards.

[...]

PR #228, opened 2024-04-11
Fix Javascript rewriting: module detection + wombat JS URL rewriting / enhance Vimeo support

### LINKED ISSUE #189: Investigate js vs module rewrite.
This issue tries to resume our problem about rewrite js as module or classic script.
For information, a document about module can be found at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

## What is different in our rewriting

In our case, with the PR #172, we rewrite `*.module.js` file differently that other `*.js` file.
When using both transformation on `*.module.js` related to vimeo we have the following differences:

### Script header

At beginning of script/module we insert some js code to mock the classical structures the script may use.

For classic js, it is:
```js
var _____WB$wombat$assign$function_____ = function(name) {return (self._wb_wombat && self._wb_wombat.local_init && self._wb_wombat.local_init(name)) || self[name]; };
if (!self.__WB_pmw) { self.__WB_pmw = function(obj) { this.__WB_source = obj; return this; } }
{
let window = _____WB$wombat$assign$function_____("window");
let globalThis = _____WB$wombat$assign$function_____("globalThis");
let self = _____WB$wombat$assign$function_____("self");
let document = _____WB$wombat$assign$function_____("document");
let location = _____WB$wombat$assign$function_____("location");
let top = _____WB$wombat$assign$function_____("top");
let parent = _____WB$wombat$assign$function_____("parent");
let frames = _____WB$wombat$assign$function_____("frames");
let opener = _____WB$wombat$assign$function_____("opener");
let arguments;
```

For module it is:
```js
import { window, globalThis, self, document, location, top, parent, frames, opener } from "../../../../_zim_static/__wb_module_decl.js";
```

The content of `__wb_module_decl.js` can be found in https://github.com/openzim/warc2zim/blob/warc2zim2/src/warc2zim/statics/__wb_module_decl.js

This is because of https://github.com/openzim/warc2zim/blob/warc2zim2/src/warc2zim/content_rewriting/js.py#L265-L266

### Inner transformation.

`import` function call are rewritten differently, a `import(foo)` will be rewritten:
- for classic js : `____wb_rewrite_import__("", t.href)`
- for module : `____wb_rewrite_import__(import.meta.url, t.href)`

This is because of https://github.com/openzim/warc2zim/blob/warc2zim2/src/warc2zim/content_rewriting/js.py#L99-L112

## How we detect it is a module

When script is imported from html, the `type="module"` attribute mark the script as being a module:
`<script type="module" src="main.js"></script>` or `<script type="module">...</script>`.
Else it is a module.

If script is also imported with a `import` (in a module only ?), it must also be used as a module.

In #172, we check the `type="module"` for inline script and if `module.js` is present in the path for item.
We also use some regex on the content to detect if it is a module: https://github.com/openzim/warc2zim/blob/warc2zim2/src/warc2zim/content_rewriting/js.py#L239-L244, but it was there before #172, so either is not enough or it is buggy.

### LINKED ISSUE #227: Zimit2: bundle/test/release javascript getWombatInfo configuration as a real dependency
As of today, configuration of wombat.js in warc2zim2 is done in the https://github.com/openzim/warc2zim/blob/warc2zim2/src/warc2zim/templates/wombat_setup.js file

This file is responsible to setup wombat:
- configure some constant settings (is_framed, is_live, static_prefix, target_frame, ...)
- but also configure the URL rewrite function which is responsible to transform URL captured by wombat into a URL which will be able to be served from the ZIM

As of today, this URL rewrite function is also implemented inside this same file. 

This function is very important since it is the central piece allowing wombat to redirect every JS network call to the proper ZIM resource.

The main limitation of current approach is that it is not possible to test this function easily. No tests are present as of today.

Not testing this function with a set of automated tests is a big concern from my PoV.

Should we consider to bundle/test/release this URL rewrite function as a real JS library (even if quite tiny) so that it will be more isolated and easier to test?

@mgautierfr @kelson42 @rgaudin any PoV?

benoit74:
Fix #189
Fix #227

This also significantly enhance the situation regarding support of Vimeo videos: #165 (but there is still some black magic from my PoV around DS rules + there is still a zimit/crawler issue to properly retrieve the whole video)

Changes:
- Move JS code in `wombat_setup.js` to a "standalone" JS codebase
  - This allows to apply traditional JS tooling on the codebase (prettier and eslint for now)
  - This allows to test and lint the codebase 
  - `wombatSetup.js` (renamed to match JS conventions) is now generated with `rollup` bundler, same bundler than webrecorder is using to bundle `wombat.js`
- Fuzzy rules are generated (via a Python script) from a JSON data source to both Python and JS code
  - The main driver was the move to a "standalone" JS codebase
  - This is however also a nice move to simplify logic (it is now way clearer where these rules are and where they are used)
  - This is also a nice move to allow to test them in the near future in JS as well (important to confirm they have the same effect in Python and in JS, since it must be kept in sync)
- Fix a significant number of bugs in the JS URL rewriting function from `wombatSetup.js`, mainly around special characters in path and in hostname
  - This also showed the need for more tests in Python codebase around URL rewriting due to edge case detected while debugging JS code
  - A very attentive reader might note that Python and JS URL rewriting are not totally aligned due to small difference between Python and JS considerations of what needs to be encoded
    - Python quotes everything but `A–Z a–z 0–9 - _ . ~`
    - JS quotes everything but `A–Z a–z 0–9 - _ . ~ ! * ' ( )`
    - e.g. a document at ZIM path `doc-um)ent.html` will be accessed from `docu-m%29ent.html` if URL is statically rewritten by Python code and from `docu-m)ent.html` if URL is dynamically rewritten by `wombat.js` and our JS rewriting function
    - We probably do not care, ZIM path is never encoded anyway, and readers are expected to be sufficiently robust to support both encoded and unencoded (or more exactly less-encoded) URLs
- JS modules are now detected by interpreting HTML and JS code rather than relying on regular expressions
  - in HTML code, `<script type="module">` and `<link rel="modulepreload">` are both supported (even if the modulepreload makes probably no sense if the script is not used somewhere after
  - in JS code, `import ...` statements are used to detect "child" JS modules
    - we assume a single JS file cannot be at the same time a module and not a module
  - Both HTML and JS rewriter are notifying that they've found a JS module based on its ZimPath
    - when notifying, the scraper store this information in a in-memory set
  - When rewriting JS, we check if current element is JS module based on the in-memory set
  - This change needed some code adaptations in HTMLRewriter so that utility function of this file have access to the current HTMLRewriter members
- Vimeo fuzzy rules have been adapted as well
  - While wabac.js are supporting Vimeo URLs like `gcs-vimeo.akamaized.net`, `vod.akamaized.net` and `vod-progressive.akamaized.net`, our test case on test-website is using `vod-adaptive.akamaized.net`
  - I've detected another issue with image placeholders on `i.vimeocdn.com` domain where resolution is dynamically adapted based on browser viewport size most probably
- Many JS test cases have been deployed on the test-website (most have already been merged before this PR)

Open issue left for later: https://github.com/openzim/warc2zim/issues/230 (wombat issue with Javascript importmap, not something we can fix on our own)

Test WARC for the Javascript part: https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/crawl-javascript-20240412.warc.gz

Test ZIMs for the Javascript part:
- ZIM before fixes: https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/javascript_not_fixed_2024-04.zim (not really interesting, but glad to see the difference ^^)
- ZIM after fixes: https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/javascript_fixed_2024-04.zim

Test WARCs for Vimeo part: 
- https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/crawl-vimeosmall-20240412.warc.gz (only few seconds of video)
- https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/crawl-vimeobig-20240412.warc.gz (most parts of the video, but still a timeout encountered in crawler)

Test ZIMs for Vimeo part: 
- https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/vimeosmall_2024-04.zim (only few seconds of video)
- https://tmp.kiwix.org/ci/javascript-and-vimeo-fixes/vimeobig_2024-04.zim (most parts of the video, but still a timeout encountered in crawler)


benoit74:
@rgaudin @mgautierfr good luck, I tried to made meaningful commits and to not break everything at once, but this is still a huge PR (but result is stunning, so definitely worth it I hope)

javascript/test/wombatUrlRewriting.js
```diff
@@ -0,0 +1,852 @@
...
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+    ),
+    undefined,
+  );
+});
+
+// Why? I don't get it ... this is not supposed to be a valid URL ...
```
mgautierfr: This is a relative url (probably not intended by user, but still handle as is). So nothing to transform ?

javascript/test/wombatUrlRewriting.js
```diff
@@ -0,0 +1,852 @@
...
+      t.context.originalHost,
+      t.context.originalScheme,
+      t.context.originalUrl,
+      t.context.prefix,
+      'https://www.example.com/javascript/content.txt?query=val%3Deue',
+      undefined,
+      undefined,
+      undefined,
+    ),
+    'http://library.kiwix.org/content/myzim_yyyy-mm/www.example.com/javascript/content.txt%3Fquery%3Dval%3Deue',
```
mgautierfr, line 474: Should we encode the original `%` (in `%3D`) ?
Else when server will decode the request path, it will decode it to `https://www.example.com/javascript/content.txt?query=val=eue` while the website is searching for `https://www.example.com/javascript/content.txt?query=val%3Deue`

javascript/src/wombatSetup.js
```diff
@@ -26,63 +27,62 @@ export function urlRewriteFunction(
...
+  // If URI scheme is defined but not http or https, we have to not rewrite the URL
+  const uri = URI.parse(url);
+  if (
+    typeof uri.scheme !== 'undefined' &&
+    !['http', 'https'].includes(uri.scheme)
+  )
+    return url;
+
+  // Compute the absolute URI, just like the browser would have resolved it hopefully
+  const original_absolute_url = URI.resolve(orig_url, url);
```
mgautierfr, line 60: Initial code resolve relative url based on `current_url`, not `orig_url`.
This is important as current url is not the same as orig url if fuzzy matching has been applied.

javascript/test/wombatUrlRewriting.js
```diff
@@ -34,24 +33,6 @@ test('undefined', (t) => {
...
-      t.context.prefix,
-      'www.example.com/javascript/content.txt',
-      undefined,
-      undefined,
-      undefined,
-    ),
-    'www.example.com/javascript/content.txt',
-  );
-});
-
```
mgautierfr: What is the output of this test with the new algorithm ?

src/warc2zim/content_rewriting/generic.py
```diff
@@ -94,6 +96,8 @@ def rewrite(
             return self.rewrite_css()
 
         if self.rewrite_mode == "javascript":
+            if any(path.value == self.path for path in self.js_modules):
```
mgautierfr, line 99: Could we implement a `__eq__` on `ZimPath` to avoid this `path.value` ?

src/warc2zim/content_rewriting/html.py
```diff
@@ -95,10 +52,25 @@ def send(self, value: str):
...
-                self._active_tag = "json"
+            script_type = self.extract_attr(attrs, "type")
+            if script_type == "json":
+                self.html_rewrite_context = "json"
+            elif script_type == "module":
+                self.html_rewrite_context = "js-module"
+            else:
+                self.html_rewrite_context = "js-classic"
+        elif tag == "link":
+            self.html_rewrite_context = "link"
```
mgautierfr, line 64: This is tricky.
You have removed the previous early `self._active_tag = tag` (the logic was by default _active_tag is tag except special case) and move to the logic "handle special case else falldown to `self.html_rewrite_context = tag` **except** here when you have a default first then special case.

It would be better to always have the same logic. Either default first and we override with special case; or if/elif/else logic, but we need a `else` all the time.

src/warc2zim/content_rewriting/html.py
```diff
@@ -119,25 +91,31 @@ def handle_starttag(self, tag: str, attrs: AttrsList, *, auto_close: bool = Fals
...
+                    extra_rules=get_ds_rules(self.url_rewriter.article_url.value),
+                    notify_js_module=self.notify_js_module,
+                ).rewrite(
+                    data,
+                    opts={"isModule": self.html_rewrite_context == "js-module"},
+                )
+        elif self.html_rewrite_context == "json":
             if data.strip():
                 rules = get_ds_rules(self.url_rewriter.article_url.value)
                 if rules:
```
mgautierfr, line 121: I would keep the simple `self.html_rewrite_context == "foo"` test together.

src/warc2zim/content_rewriting/html.py
```diff
@@ -161,3 +139,50 @@ def handle_pi(self, data: str):
...
+        processed_attrs = (self.process_attr(attr, url_rewriter) for attr in attrs)
+        return " ".join(self.format_attr(*attr) for attr in processed_attrs)
+
+    def extract_attr(
+        self, attrs: AttrsList, name: str, default: str | None = None
+    ) -> str | None:
+        for attr_name, value in attrs:
+            if attr_name == name:
+                return value
+        return default
```
mgautierfr, line 188: You don't help us to review this change by moving to function code elsewhere and updating it in the same time. Moving the free function as a method and then updating it would have been simpler.

rules/rules.json
```diff
@@ -21,8 +21,8 @@
             "replace": "youtube.fuzzy.replayweb.page/embed/\\1"
         },
         {
-            "pattern": ".*(?:gcs-vimeo|vod|vod-progressive)\\.akamaized\\.net.*?/([\\d/]+.mp4)$",
-            "replace": "vimeo-cdn.fuzzy.replayweb.page/\\1"
+            "pattern": ".*(?:gcs-vimeo|vod|vod-progressive|vod-adaptive)\\.akamaized\\.net.*/(.+?.mp4)\\?.*range=(.*?)(?:&|$)",
+            "replace": "vimeo-cdn.fuzzy.replayweb.page/\\1?range=\\2"
```
mgautierfr, line 25: Where this change come from ?

You find it or it is coming from wabac update ?

javascript/src/wombatSetup.js
```diff
@@ -30,6 +30,20 @@ export function urlRewriteFunction(
...
+    typeof uri.scheme !== 'undefined' &&
+    !['http', 'https'].includes(uri.scheme)
+  )
+    return url;
+
+  // If url starts with prefix, we need to remove this prefix before applying usual
+  // rewrite rules
+  if (url.startsWith(prefix)) {
+    url = uri.scheme + '://' + url.substring(prefix.length);
+  }
```
mgautierfr, line 45: ~~Do we really need to apply rewrite rules if url is already rewritten ?~~

We actually need to as we need to properly encode it. Maybe we could separate the encoding from the rewriting (relative path and fuzzy) ?

rules/rules.json
```diff
@@ -27,6 +27,10 @@
         {
             "pattern": ".*player.vimeo.com/(video/[\\d]+)\\?.*",
             "replace": "vimeo.fuzzy.replayweb.page/\\1"
+        },
+        {
+            "pattern": ".*i\\.vimeocdn\\.com\\/(.*)\\?.*",
+            "replace": "i.vimeocdn.fuzzy.replayweb.page/\\1"
```
mgautierfr, line 33: Maybe we could send those rules update upstream ?

mgautierfr (changes requested):
Impressive work. I have few remarks but mostly good.

I think your solution (registering js module as we found them in the same time we rewrite content) works because we handle content in the same order it as been recorded (at browsertrix load js after they are "requested" in parsed content).

It think it would be good to mention this somewhere.

benoit74:
> I think your solution (registering js module as we found them in the same time we rewrite content) works because we handle content in the same order it as been recorded (at browsertrix load js after they are "requested" in parsed content).
>
> It think it would be good to mention this somewhere.

I thought I did, but it looks like you did not found it and I do not find it either now, so ... But anyway, yes, you are right.

rules/rules.json
benoit74: Definitely, I just wanted to have your feedback first.

javascript/src/wombatSetup.js
benoit74: I don't mind to separate these two parts, but I do not consider there will be a great benefit, even if we consider testing. So why not, but not urgent at all (i.e. will probably never be done AFAIK ^^)

rules/rules.json
benoit74: I found it. To be reported upstream as well.

src/warc2zim/content_rewriting/html.py
benoit74: Agreed

src/warc2zim/content_rewriting/html.py
benoit74: Agreed, sorry for that. Do you want that I make this move in another PR first? (but then I will have to force-push in this PR branch to do the rebase)

src/warc2zim/content_rewriting/html.py
benoit74: You mean, push the `self.html_rewrite_context and self.html_rewrite_context.startswith("js-")` to the end?

src/warc2zim/content_rewriting/generic.py
benoit74: We could, but I don't want to ^^

Being forced to do it is exactly the purpose of this class (and the `HttpUrl`) so that we make it intentional to compare a `ZimPath` with a string (should not be the normal case).

Actually here the mistake is that I should pass a `ZimPath` to the rewriters to make things even clearer. But this is beyond the scope of this PR, I will move this to a dedicated issue: https://github.com/openzim/warc2zim/issues/236

benoit74:
@rgaudin please review the code as-is, in fact there isn't much changes expected for now and changing them now could even make your review harder.

docs/technical_architecture.md
```diff
@@ -1,6 +1,24 @@
 # Technical architecture
 
-## High level overview
+## Fuzzy rules
+
+Fuzzy rules are stored in `rules/rules.json`. This configuration file is then used by `rules/generateRules.py` to
```
rgaudin, line 5: Typo ? see next line

javascript/src/wombatSetup.js
```diff
@@ -0,0 +1,206 @@
...
+  if (
+    typeof uri.scheme !== 'undefined' &&
+    !['http', 'https'].includes(uri.scheme)
+  )
+    return url;
+
+  // If url starts with prefix, we need to remove this prefix before applying usual
+  // rewrite rules
+  if (url.startsWith(prefix)) {
+    url = uri.scheme + '://' + url.substring(prefix.length);
```
rgaudin, line 44: Can we mention above (in prefix comment) what prefix looks like.

javascript/src/wombatSetup.js
rgaudin: @benoit74 was this intentional? If yes why and comment

javascript/src/wombatSetup.js
```diff
@@ -0,0 +1,206 @@
...
+    url = encodeURI(url);
+  }
+
+  // Compute the absolute URI, just like the browser would have resolved it hopefully
+  const original_absolute_url = URI.resolve(orig_url, url);
+
+  // We now have to transform this absolute URI into a normalized ZIM path entry
+  const absolute_url_parts = URI.parse(original_absolute_url);
+
+  // Let's first compute the decode host
```
rgaudin, line 65:
```suggestion
  // Let's first compute the decoded host
```

javascript/src/wombatSetup.js
```diff
@@ -0,0 +1,206 @@
...
+  );
+  const decoded_host = serialized_host.substring(7, serialized_host.length - 1);
+
+  // And the decoded path, only exception is that an empty path must resolve to '/' path
+  const decoded_path =
+    !absolute_url_parts.path || absolute_url_parts.path.length === 0
+      ? '/'
+      : decodeURIComponent(absolute_url_parts.path);
+
+  // And the decoded query, only exception is that + sign must resolve to ' ' to avoid confusion
```
rgaudin, line 78: Why is that? Our convention? Wombat?

javascript/src/wombatSetup.js
```diff
@@ -0,0 +1,206 @@
...
+
+  // apply the fuzzy rules to the ZIM path
+  const fuzzifiedPath = applyFuzzyRules(zimPath);
+
+  // Reencode everything but '/' (we decode it afterwards for simplicity)
+  const finalUrl =
+    prefix + encodeURIComponent(fuzzifiedPath).replaceAll('%2F', '/');
+
+  /*
+  console.log(
```
rgaudin, line 95:
```suggestion
  console.debug(
```

javascript/src/wombatSetup.js
```diff
@@ -0,0 +1,206 @@
...
+    proxy_magic: '',
+
+    // This is the prefix on which we have stored our static files (needed by wombat).
+    // Must not conflict with other url served.
+    // Will be used by wombat to not rewrite back the url
+    static_prefix: prefix + '_zim_static/',
+
+    wombat_ts: '',
+
+    // A delay is sec to apply to all js time (`Date.now()`, ...)
```
rgaudin, line 182:
```suggestion
    // A delay in sec to apply to all js time (`Date.now()`, ...)
```

javascript/src/wombatSetup.js
```diff
@@ -0,0 +1,206 @@
...
+    wombat_host: orig_host,
+
+    // Extra options ?
+    wombat_opts: {},
+
+    // ?
+    enable_auto_fetch: true,
+    convert_post_to_get: true,
+    target_frame: '___wb_replay_top_frame',
+    isSW: true,
```
rgaudin, line 198: What's this and why should it be true?

rules/generate_rules.py
```diff
@@ -0,0 +1,71 @@
+import json
+import re
+import sys
+from pathlib import Path
+
+from jinja2 import Environment
+
+rules_src = Path(__file__).parent / "rules.json"
```
rgaudin, line 8:
```suggestion
rules_src = Path(__file__).with_name("rules.json")
```

rules/generate_rules.py
```diff
@@ -0,0 +1,71 @@
...
+if not rules_src.exists():
```
rgaudin, line 9: What's the use case ? Why not failing?

rules/generate_rules.py
```diff
@@ -0,0 +1,71 @@
...
+    # This skip is usefull mostly for CI operations when installing only Python deps
+    print("Skipping rules generation, rule file is missing")
+    sys.exit()
+
+FUZZY_RULES = json.loads((Path(__file__).parent / "rules.json").read_text())[
```
rgaudin, line 14:
```suggestion
FUZZY_RULES = json.loads(rules_src.read_text())[
```

rules/generate_rules.py
```diff
@@ -0,0 +1,71 @@
...
+{% for rule in FUZZY_RULES %}  {
+    match: '{{ rule['match'] }}',
+    replace: '{{ rule['replace'] }}',
+  },
+{% endfor %}
+];
+
+"""
+
+js_parent = Path(__file__).parent.parent / "javascript" / "src"
```
rgaudin, line 36: I find this more readable… but maybe not windows compat
```suggestion
js_parent = Path(__file__).joinpath("../javascript/src").resolve()
```

rules/generate_rules.py
```diff
@@ -0,0 +1,71 @@
...
+if not js_parent.exists():
+    # This skip is usefull mostly for CI operations when working on the Python part
+    print("Skipping JS rules generation, target folder is missing")
+else:
+    (js_parent / "fuzzyRules.js").absolute().write_text(
```
rgaudin, line 41: Why the `.absolute()`?

tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+
+
+@pytest.fixture(scope="module")
```
rgaudin, line 4: docstring

src/warc2zim/url_rewriting.py
```diff
@@ -285,6 +253,13 @@ def __init__(self, article_url: HttpUrl, existing_zim_paths: set[ZimPath]):
         self.existing_zim_paths = existing_zim_paths
         self.missing_zim_paths: set[ZimPath] = set()
 
+    def get_item_path(self, item_url: str) -> ZimPath:
+        """Utility to transform an item URL into a ZimPath"""
+
+        item_absolute_url = urljoin(self.article_url.value, item_url)
+        item_path = normalize(HttpUrl(item_absolute_url))
```
rgaudin, line 260:
```suggestion
        return = normalize(HttpUrl(item_absolute_url))
```

src/warc2zim/content_rewriting/html.py
```diff
@@ -95,10 +52,25 @@ def send(self, value: str):
         self.output.write(value)  # pyright: ignore[reportOptionalMemberAccess]
 
     def handle_starttag(self, tag: str, attrs: AttrsList, *, auto_close: bool = False):
-        self._active_tag = tag
         if tag == "script":
-            if "json" in (extract_attr(attrs, "type") or ""):
-                self._active_tag = "json"
+            script_type = self.extract_attr(attrs, "type")
+            if script_type == "json":
```
rgaudin, line 57:
```suggestion
            self.html_rewrite_content = {"json": "json", "module": "js-module"}.get(script_type, "js-classic")
```

src/warc2zim/content_rewriting/html.py
```diff
@@ -161,3 +139,50 @@ def handle_pi(self, data: str):
 
     def unknown_decl(self, data: str):
         self.handle_decl(data)
+
+    def process_attr(
+        self,
+        attr: tuple[str, str | None],
```
rgaudin, line 145: What is expected value? Since we don't expand

src/warc2zim/content_rewriting/html.py
```diff
@@ -161,3 +139,50 @@ def handle_pi(self, data: str):
 
     def unknown_decl(self, data: str):
         self.handle_decl(data)
+
+    def process_attr(
```
rgaudin, line 143: no docstring, using indexed values without any information… that's not maintenance-friendly 😉

rgaudin (commented):
Thank you ; I've noted somme minor suggestions.

javascript/src/wombatSetup.js
benoit74: This is intentional, but I agree there is still a bug. I'm looking into it right now. 

Using `orig_url` is mandatory to properly rewrite the URL when it is an absolute URL on a "new" hostname and we want to properly encode the hostname. Maybe the trick is to make a difference between the case where we have an absolute URL and the case where it is a relative one (which might hence be already rewritten, we just don't know).

benoit74: I will try to summarize the bug (I'm not sure yet on how to solve it).

Let's say:
- the ZIM is served at `https://mylibrary.com/content/myzim/`
- we are on a page originally located at `https://www.example.com/index.html`
- Wombat passes us the following URL to rewrite: `../www.otherdomain.com/image.png`. 

I don't know how to differentiate two cases with 100% certainty.

Case 1: 
- the URL has not been rewritten yet 
- the resource was originally located at `https://www.example.com/www.otherdomain.com/image.png`
- the rewritten URL must be `https://mylibrary.com/content/myzim/www.example.com/www.otherdomain.com/image.png`
- there was a silent bug in HTML code where the relative link was going "too high" in the tree, but AFAIK most browser are very permissive with this scenario

Case 2:
- the URL has already been rewritten
- the resource was originally located at `https://www.otherdomain.com/image.png`
- the rewritten URL must be `https://mylibrary.com/content/myzim/www.otherdomain.com/image.png`

A dirty hack would be to include a magic querystring (or prefix, or ...) in all rewritten URL so that we are 100% percent sure in which scenario we are.

Another solution would be to use a heuristic: if we are rewriting a relative link AND there is at least one dot in the first path segment (indicating a probably hostname, quite rare in directories) AND the link is going exactly one level too high, then we can be almost 100% confident the URL to rewrite has indeed already been rewritten and we can keep it as-is.

As you can see, using `current_url` instead of `orig_url` does not help at all, it just assumes that we are always in case 2 by allowing JS to go one level higher when resolving the `../`.

I would prefer to implement the heuristic, but did I missed a more straightforward solution or am I wrong somewhere? Any idea?

src/warc2zim/content_rewriting/html.py
mgautierfr: yes

.gitignore
```diff
@@ -206,14 +471,40 @@ poetry.toml
...
-tmp
\ No newline at end of file
+tmp
+
+# rule files are generated by rules/generate_rules.py
+src/warc2zim/rules.py
+javascript/src/fuzzyRules.js
+
+# wombatSetup.js is generated with rollup
+src/warc2zim/statics/wombatSetup.js
```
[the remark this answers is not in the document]
benoit74: I like your arguments. I proposed to keep it as it is for now since this is how we've already done on most Python scrapers. I still think current approach is wrong and your arguments are correct, so I've opened https://github.com/openzim/_python-bootstrap/issues/41 so that we track the decision + document what needs to be.

javascript/test/wombatUrlRewriting.js
benoit74: In python we are not encoding the original `%`, we keep it in fully decoded form inside the ZIM path. I've done the same in Javascript.

javascript/test/wombatUrlRewriting.js
benoit74: It should be transformed as well, this is why I don't get why it was not transformed previously

javascript/src/wombatSetup.js
benoit74: We've discussed it live with @mgautierfr 

We've chosen a middle ground:
- rewrite all domain names by appending a `.zim` suffix 
- use that to detect that a relative URL has already been rewritten
- document that domains names ending with `.zim` are not supported (this TLD does not yet exists)

docs/technical_architecture.md
benoit74: Solved

javascript/src/wombatSetup.js
benoit74: I've added sample value for all args indeed ^^

rules/rules.json
benoit74: Done: https://github.com/webrecorder/wabac.js/issues/169

javascript/src/wombatSetup.js
benoit74: I've opened https://github.com/openzim/warc2zim/issues/239

rules/generate_rules.py
benoit74: It's needed for the "first pass" of Docker build, where rules are not yet available / we do not want to generate this because we do not have yet the proper folder structure. Pretty harmless anyway

rules/generate_rules.py
benoit74: I don't know ... not needed indeed

src/warc2zim/content_rewriting/html.py
benoit74: I don't get your question, sorry

benoit74: Nevermind, I got it.

benoit74: Now it is expanded ^^

src/warc2zim/content_rewriting/html.py
benoit74: Same for all functions in this file. I've tracked this in https://github.com/openzim/warc2zim/issues/219

javascript/src/wombatSetup.js
benoit74: Edit: rewriting all domain names by appending a `.zim` suffix does not helped at all indeed. We still need to detect the difference between a real path ending with `.zim` and a "fake / ZIM-only" path ending with `.zim` because it is the domain which has been encoded there.

I hence finally implemented the heuristic I spoke about in previous comment, implemented a test-case on the test website and everything is working fine now (while it was broken before).

benoit74:
@mgautierfr @rgaudin this is ready for a new review. I still hope it is not going to be too painful, but I know the truth is probably harder than that. As usual, I've left conversations opened when I consider you might wanna check what has been done since your last review. I hope I did not forgot anything.

Tell me if it is just too painful and you consider you might merge and see what's going on in real test cases.

rgaudin (approved):
Thank you ; easier to read with those improvements. Spotted a couple of minor stuff that you should look into but LGTM.

javascript/src/wombatSetup.js
```diff
@@ -15,12 +15,12 @@ export function applyFuzzyRules(path) {
 }
 
 export function urlRewriteFunction(
-  current_url, // The current (real) url we are on
-  orig_host, // The host of the original url
-  orig_scheme, // The scheme of the original url
-  orig_url, // The original url
-  prefix, // The (absolute) prefix to add to all our urls (from where we are served))
-  url, // first argument passed by wombat.JS at each invocation
+  current_url, // The current (real) url we are on, e.g. http://library.kiwix.org/content/www.example.com/index.html
```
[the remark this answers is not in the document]
benoit74: This is not **very unlikely**, this is simply wrong ... both ZIM name and domain name are always present inside the path so this example value will never exist. Thank you very much for spotting this.

javascript/src/wombatSetup.js
```diff
@@ -57,31 +57,72 @@ export function urlRewriteFunction(
   }
 
   // Compute the absolute URI, just like the browser would have resolved it hopefully
+  // We need to use the original URL for that to properly detect the hostname when
+  // present ; current URL does not allow to do it easily
   const original_absolute_url = URI.resolve(orig_url, url);
 
+  // Detect (with a heuristic) that the path is most probably already rewritten and
```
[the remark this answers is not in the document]
benoit74: Done

rules/generate_rules.py
```diff
@@ -33,12 +31,13 @@
 
 """
 
-js_parent = Path(__file__).parent.parent / "javascript" / "src"
+js_parent = Path(__file__).joinpath("../../javascript/src").resolve()
+print(js_parent)
```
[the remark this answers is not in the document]
benoit74: Yep

benoit74:
renaud comments have been fixed

src/warc2zim/url_rewriting.py
```diff
@@ -280,7 +285,10 @@ def __call__(self, item_url: str, *, rewrite_all_url: bool = True) -> str:
         if rewrite_all_url or item_path in self.existing_zim_paths:
             return self.get_document_uri(item_path, item_fragment)
         else:
-            if item_path not in self.missing_zim_paths:
+            if (
+                isinstance(self.missing_zim_paths, set)
```
mgautierfr, line 288: `self.missing_zim_paths is not None` is probably better than `instance(self.missing_zim_paths, set)`

mgautierfr (approved):
Just a small comment on code. Mostly a matter of style.
Will not block for that.

Else we are good so it is an approval.

PR #233, opened 2024-04-17
Add debug logging for converter properties

richterdavid:
Initial take on addressing issue 123, for feedback.


richterdavid:
I'd like feedback on the approach, formatting, and whether logged property names should reflect the variable names in the code or the property names in https://wiki.openzim.org/wiki/Metadata. If the latter, I may need some guidance on the mapping. For example, I suspect that `main_url` maps to `Source`, but this isn't made clear in the source code.

benoit74:
Closing it, nothing will have to be done at warc2zim level, see https://github.com/openzim/python-scraperlib/issues/155

PR #249, opened 2024-05-13
No more failure due to url rewriting issue

### LINKED ISSUE #247: Warc2zim2 is not sufficiently resilient to statical URL rewrite issues
In warc2zim, we rewrite URLs statically in documents (HTML, JS, CSS, ...) so they point to the right ZIM entry.

While this is a mandatory action for proper operation, it might fail on some occasion especially when the source document is incorrect.

For instance, following HTML code is impossible to rewrite properly: `<a href="http:"></a>`

Currently, when such a situation occurs the scraper is stopped with a failure.

This is not correct, since we know that a bad source document cannot be recovered at all (or at least we do not want to).

As a consequence, we need to ensure that any URL rewriting issue is just raising a warning (probably only in DEBUG level) inside the log but not failing the WARC conversion or the WARC item conversion. We just need to keep the URL as-is.

benoit74:
## Rationale

Fix #247 

## Changes

- all URL rewriting issues are simply logged, but they do not cause the scraper to fail
- no more special handling of hostname rewriting issues
  - try to decode hostname only if it starts with "xn--" (the indicator used for IDNA puny-encoded see [here](https://web.archive.org/web/20100427154004/http://www.atm.tut.fi/list-archive/ietf-announce/msg13572.html))

### DIFF
--- a/src/warc2zim/url_rewriting.py
+++ b/src/warc2zim/url_rewriting.py
@@ -276,32 +276,55 @@ class ArticleUrlRewriter:
         The url is "fully" rewrited to point to a normalized entry path
         """
 
-        item_url = item_url.strip()
+        try:
+            item_url = item_url.strip()
 
-        item_scheme = urlsplit(item_url).scheme
-        if item_scheme and item_scheme not in ("http", "https"):
+            item_scheme = urlsplit(item_url).scheme
+            if item_scheme and item_scheme not in ("http", "https"):
+                return item_url
+
+            item_absolute_url = urljoin(
+                urljoin(self.article_url.value, base_href), item_url
+            )
+
+            item_fragment = urlsplit(item_absolute_url).fragment
+
+            item_path = normalize(HttpUrl(item_absolute_url))
+
+            if rewrite_all_url or item_path in self.existing_zim_paths:
+                return self.get_document_uri(item_path, item_fragment)
+            else:
+                if (
+                    self.missing_zim_paths is not None
+                    and item_path not in self.missing_zim_paths
+                ):
+                    logger.debug(f"WARNING {item_path} ({item_url}) not in archive.")
+                    # maintain a collection of missing Zim Path to not fill the logs
+                    # with duplicate messages
+                    self.missing_zim_paths.add(item_path)
+                # The url doesn't point to a known entry
+                return item_absolute_url
+
+        except Exception as exc:
+            item_scheme = item_scheme if "item_scheme" in locals() else "<not_set>"
+            item_absolute_url = (
+                item_absolute_url if "item_absolute_url" in locals() else "<not_set>"
+            )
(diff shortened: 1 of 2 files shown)


src/warc2zim/url_rewriting.py
```diff
@@ -276,32 +271,58 @@ def __call__(
...
+                ):
+                    logger.debug(f"WARNING {item_path} ({item_url}) not in archive.")
+                    # maintain a collection of missing Zim Path to not fill the logs
+                    # with duplicate messages
+                    self.missing_zim_paths.add(item_path)
+                # The url doesn't point to a known entry
+                return item_absolute_url
+
+        except Exception as exc:
+            if logger.isEnabledFor(logging.DEBUG):
```
rgaudin, line 304: There's no heavy processing here so running this conditionally isn't justified

src/warc2zim/url_rewriting.py
```diff
@@ -194,18 +192,15 @@ def normalize(url: HttpUrl) -> ZimPath:
...
-            hostname = idna.decode(hostname)
-        except idna.IDNAError as exc:
-            # exception might happen if illegal character are found in hostname like the
-            # `_` in `host_ip` ; we keep the set of bad hostname in memory to not fill
-            # the logs with these warnings
-            logger.warning(f"Bad hostname found, kept as-is: {hostname}", exc_info=exc)
-            known_bad_hostnames.add(hostname)
+    # decode the hostname if it is punny-encoded
+    hostname = (
+        idna.decode(url_parts.hostname)
```
rgaudin, line 200: With this we're not rewriting at all in case of IDNA parsing issue (on xn-- prefixed domain).
In such case, don't we want to still rewrite (reusing whatever bad domain there is)?

rgaudin (changes requested):

src/warc2zim/url_rewriting.py
benoit74: Agreed !

src/warc2zim/url_rewriting.py
benoit74: As discussed live, the idea is that we do not trap any exception at this level because:
- if we are normalizing the URL from a WARC record, then hostname should be valid (browser achieved to fetch it) ; if it fails, then computing a badly formatted ZIM path is anyway error-prone and probably doomed to fail, so we let the error raise and it is catched upper in the stack and stops the scraper (and we will be able to revisit why this happened and fix this edge-case)
- if we are normalizing a URL found in a document, then when hostname is invalid then we know we will most probably not find a corresponding ZIM entry, so we let the error raise and it is catched upper in the stack and conducts the URL to be kept as-is

benoit74:
Merging right away as this has been discussed live with rgaudin

PR #251, opened 2024-05-13
Properly notify JS module location when a base href is configured

### LINKED ISSUE #250: base href is not properly handled for notification of JS modules
JS code is already properly rewritten when a base href is configured in the HTML document, but the notification system used to know which JS file is a module is not properly taking into account the base href from the HTML document.

For JS modules, the base href of HTML document must be taken into account to properly notify of the module location (both in external script and when using inline JS module).

benoit74:
Fix #250


rgaudin (approved):
LGTM

PR #252, opened 2024-05-13
Simplify debugging of WARC record processsing issues

### LINKED ISSUE #248: Enhance ability to debug WARC item conversion issues
Currently, when the scraper fails to process a WARC record, the scraper immediately stops. 

Since this might happen after few minutes of processing (we have some huge WARC files to process), it both quite painful for the developer, especially since only logs are available after that.

We could consider to add an option to:
- try/catch every WARC records processing
- in DEBUG mode, add more logs when a WARC record fails (record headers, HTTP headers) and save record content base64 encoded in the output folder
- add a CLI flag "--continue-on-error" and process accordingly (to avoid fixing issues one-by-one) 

This option should not be exposed in the Zimfarm.

benoit74:
Fix #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,14 @@ warc2zim myarchive.warc --name myarchive -i subdomain1.example.com -i subdomain2
 
 `--custom-css` allows passing an URL or a path to a CSS file that gets added to the ZIM and gets included on **every HTML article** at the very end of `</head>` (if it exists).
 
+### Failed items
+
+When an item fails to be converted into the ZIM and `--verbose` flag is passed, the failed item content is stored on the filesystem for easier analysis. The directory where this file is saved can be customized with `--failed-items`. File name is a random UUID4 which is output in the logs.
+
+### Development features
+
+For developement purpose, it is possible to ask to continue on WARC record processing errors with `--continue-on-error`.
+
 ### Other options
 
 See `warc2zim -h` for other options.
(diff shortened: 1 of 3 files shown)


rgaudin (changes requested):
Useful dev feature!

- it's a dev feature that should be advertised as such.
- it should be properly documented. Usage doesn't mention that failed records are stored on the filesystem.
- why storing the data as base64 on the filesystem? base64 is a transport-over-text system. I don't see any reason to use it here. Am I missing something?

benoit74:
> it's a dev feature that should be advertised as such.

Agreed. 

> it should be properly documented. Usage doesn't mention that failed records are stored on the filesystem.

Agreed, I will indeed even add a CLI argument to specify where one wants to store these failed items (with same default as currently, but having it is both useful in some occasions + a great way to make the feature more visible)

> why storing the data as base64 on the filesystem? base64 is a transport-over-text format. I don't see any reason to use it here. Am I missing something?

Because at first I intended to log this content, so base64 was mostly mandatory to avoid issues. I then realized it was too risky / complex / cumbersome to push potentially big content (e.g. pdfs, ...) to the log and decided to store them on the filesystem. But I didn't recall that base64 was then not needed anymore. Anyway, I agree we should store the item as-is, in "plain".

benoit74:
Sorry, small typo found in documentation after asking for review.

benoit74:
@rgaudin up, rewiew required please

rgaudin (approved):
LGTM

Issue #254, opened 2024-05-14
Processing of WARC records with HTTP status code 0


benoit74:
When implementing https://github.com/openzim/warc2zim/issues/220, we considered that HTTP status code 0 is not processable. We even had to manually edit a WARC used in the test set to alter its HTTP status code which was 0 (we considered it was an old bug).

This in fact created a regression in warc2zim, i.e. WARC record with status code 0 are not that unusual / still produced by the crawler.

See https://github.com/webrecorder/browsertrix-crawler/issues/570 for a discussion on this topic.

Until things get clear on crawler side, we obviously should consider that HTTP status code 0 is equivalent to HTTP status 200.

benoit74:
Upstream confirmed this was indeed a bug, which has been fixed. So we should definitely not consider a status code 0 as normal. I will close the PR without merging it, and create another one to just better log issues of unexpected status codes. Currently we do not make a difference between unprocessable status code (which are "normal", e.g. 404) and unexpected status codes (which are "abnormal", e.g. 0, invalid status codes, ...). Both are logged only in DEBUG log level. This is probably fine for unprocessable status code, but unexpected status codes should be logged at least in WARNING log level since they are not expected.

benoit74:
Fixed by https://github.com/openzim/warc2zim/pull/256

PR #255, opened 2024-05-14
Consider WARC record with status code 0 as normal

benoit74:
Interim solution for #254 (maybe the final fix, depending on upstream decision)

Changes:
- WARC record with status code 0 are considered as "normal" (i.e. processed just like 200, 201, ...)


rgaudin (approved):
LGTM

benoit74:
Won't merge in fact, status code 0 is not normal but an upstream bug, see https://github.com/openzim/warc2zim/issues/254#issuecomment-2114230714

PR #259, opened 2024-05-17
Document and test limitations around processing of HTML attributes

### LINKED ISSUE #219: Zimit2: Is there an HTML parser issue with some special characters?

benoit74:
Fix #219 (only documentation and tests needed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,13 @@ that can render its content in a modern browser.
   - in the `2xx` range, only `200`, `201`, `202` and `203` are supported ; others are simply ignored
   - in the `3xx` range, only `301`, `302`, `306` and `307` are supported if they redirect to a payload which is present in the WARC ; others are simply ignored
   - all payloads with HTTP return codes in the `1xx` (not supposed to exist in WARC files anyway), `4xx` and `5xx` ranges are ignored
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
LGTM

PR #266, opened 2024-05-21
Validate ZIM metadata early

### LINKED ISSUE #235: Validate metadata as early as possible
Scraper must validate metadata and fail as early as possible (especially when called without WARCs to validate inputs).

benoit74:
Fix #235 

Changes:
- Validate title, description, long_description and tags early, including during the pre-check call from zimit
- Add `--disable-metadata-check` CLI argument, not intended to be exposed in Zimfarm


rgaudin (approved):
LGTM

PR #267, opened 2024-05-21
Tags csv

### LINKED ISSUE #265: Move from multiple `--tags` flags to one single `--tags` flag
Currently, when the user wants to specify multiple ZIM tags, he has to specify multiple time the `--tag` parameter like in `warc2zim --tags "Quality Assurance" --tags "Another TAG"`

This kind of thing is not supported by the Zimfarm and is not in line with other scrapers behavior.

We hence want to move to one single `--tags` flag which might contain a CSV value like in `warc2zim --tags "Quality Assurance,Another TAG"`. Commas are not expected to be used in tag values.

NOTA: this is a breaking API change ... so better done in 2.0.0

benoit74:
Fix #265


benoit74:
Nota: depends on https://github.com/openzim/warc2zim/pull/266 that must be merged first before reviewing this PR

benoit74:
This is ready for review once #266 is merged (i.e. you can look only at last commit of the list)

src/warc2zim/converter.py
```diff
@@ -109,7 +115,20 @@ def __init__(self, args):
...
         self.publisher = args.publisher
-        self.tags = DEFAULT_TAGS + (args.tags or [])
+        self.tags = [
+            tag
+            for tag in list(
+                set(
+                    DEFAULT_TAGS
+                    + [
+                        t2.strip()
+                        for t1 in args.tags.split(",")  # support comma
```
rgaudin, line 125: We already agreed on not supporting multiple formats. Serves no purpose and brings confusion.
We also agreed to use semi colons, as it's what is stored in Metadata so we only ever support semicolon separated values.

rgaudin (changes requested):

src/warc2zim/converter.py
benoit74: That's a great decision, I unfortunately found at least two scrapers which where support only the comma and I felt it was awkward since Metadata support only semicolon. Should be shared at zimscraperlib level in the end.

benoit74:
Nota: I force-pushed a rewritten commit because there were too many changes, made no sense to review a fixup commit which was harder to read that the rewritten commit ^^

rgaudin (approved):
LGTM

PR #268, opened 2024-05-21
Fix path typing in `WARCPayloadItem` and `Rewriter`

### LINKED ISSUE #236: `WARCPayloadItem` and `Rewriter`s must have a `ZimPath` object as path instead of a string
`WARCPayloadItem` and `Rewriter`s have a `path` member which is a string. In fact, it should be kept as a `ZimPath` in these classes as well for code clarity.

benoit74:
Fix #236


rgaudin (approved):
Makes sense!

PR #269, opened 2024-05-21
Document scraper capabilities and limitations

### LINKED ISSUE #214: Zimit2: Document what is known (or supposed) to work and known limitations (or not tested at least)
It is becoming more and more important to document what is known to work and what is not to not be supported (yet or at all).

I propose a first list, @mgautierfr your help is welcomed on this

What is supported:
- ??? (I struggle to have a clear wording of what is currently supported)

What is not yet supported (but should be soon as part of Zimit2):
- pages and resources whose URLs have characters which should be encoded to be a valid URL according to W3C specification
- pages and resources whose URLs have a querystring: the querystring is more or less ignored for now, it might work if the querystring is not important (since it is usually dropped by the reader)
- pages and resources on domain name with characters which should be punny-encoded: if the 
- Javascript code doing string manipulation to compute URLs: these URLs are not rewritten and there is a significant chance they won't work
- Javascript modules: only partial support for now

What is not supported:
- mix of HTTP and HTTPS with two different resources at same hostname and path: scraper takes the simplification this situation does not exist, and does not take into account the protocol when computing ZIM entry path. E.g. if http://www.acme.com/resource1 and https://www.acme.com/resource1 both exist AND lead to different resources, the scraper will fail
- replay of HTTP response headers: these headers are not stored inside the ZIM / not replayed
- HTTP request headers: if two different request header values leads to two different page / resource, scraper is ignoring this information
- User-Agent: corollary of the point above on HTTP request headers, scraper supposes a single User-Agent has been used to create the WARC files ; if the website is providing different content based on the User-Agent, only one will be used

What I don't know at all:
- HTTP/2: ?

benoit74:
Fix #214

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,34 @@ warc2zim provides a way to convert WARC files to ZIM, storing the WARC payload a
 Additionally, the [ReplayWeb.page](https://replayweb.page) is also added to the ZIM, creating a self-contained ZIM
 that can render its content in a modern browser.
 
+## Capabilities
+
+While we would like to support as many websites as possible, making an offline archive of a website obviously has some limitations.
+
+Scenario which are known to work well:
+- HTML and CSS documents
+- JS manipulating the DOM and/or doing simple fetch (preferably GET) requests
+  - E.g. JS manipulating the DOM to modify images, fetch remote stuff (JSON data, ...) is supposed to work
+  - POST requests support is fairly limited (at best, scraper replays the same response as it has been recorded)
+- Puny-encoded hostnames
+- Encoded URL path
+- URL query string
+- URL fragments
+- JS modules
+- HTML base href
+
 ## Known limitations
 
+- Any web site expecting a server to store live data and wanting to modifying those data (form, read/write api, ...) is not supported
+- Website using dynamic resources (dynamic URLs) fetch based on user-agent configuration (e.g. viewport), timestamp, unique ID
+  - E.g. if the viewport size is sent in every requests to fetch website images, this will not work since the URL built during the scrape will most likely be different than the URL built when the end-user read the ZIM content, and the ZIM reader won't find associated resource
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,8 +12,34 @@ warc2zim provides a way to convert WARC files to ZIM, storing the WARC payload a
...
+- URL fragments
+- JS modules
+- HTML base href
+
 ## Known limitations
 
+- Any web site expecting a server to store live data and wanting to modifying those data (form, read/write api, ...) is not supported
+- Website using dynamic resources (dynamic URLs) fetch based on user-agent configuration (e.g. viewport), timestamp, unique ID
+  - E.g. if the viewport size is sent in every requests to fetch website images, this will not work since the URL built during the scrape will most likely be different than the URL built when the end-user read the ZIM content, and the ZIM reader won't find associated resource
+  - Scraper tries to do its best on few popular websites (e.g. Youtube) with fuzzy rules, but support is fairly very limited
```
rgaudin, line 36: This statement has limited value to end user. What does this fuzzy mean? Problem/solution

README.md
```diff
@@ -12,8 +12,34 @@ warc2zim provides a way to convert WARC files to ZIM, storing the WARC payload a
...
+- For simplification, scraper assumes that:
+  - servers do not mix multiple ports with two different resources at same hostname and path. E.g. if `http://www.acme.com:80/resource1` and `http://www.acme.com:8080/resource1` both exist AND lead to different resources, the scraper will fail
+  - corollary: servers do not mix HTTP and HTTPS with two different resources at same hostname and path. E.g. if `http://www.acme.com/resource1` and `https://www.acme.com/resource1` both exist AND lead to different resources, the scraper will fail
```
rgaudin, line 39: Will it fail or silently ignore one?

rgaudin (approved):
Good

README.md
benoit74: It is silently ignored indeed

PR #270, opened 2024-05-21
Add support for onxxx HTML events

### LINKED ISSUE #209: onxxx link not rewritten
Haven't investigated enough to understand which component is responsible but to reproduce:

- Go to https://dev.library.kiwix.org/viewer#solidarite-numerique_fr_2023-03/
- Click on any of the cards

You must be **using Chrome**.

This leads to the broken content Chrome response with the following message in the console:

```
Refused to frame 'https://www.solidarite-numerique.fr/' because it violates the following Content Security Policy directive: "default-src 'self' data: blob: mediastream: ws: wss:". Note that 'frame-src' was not explicitly set, so 'default-src' is used as a fallback.
```

What happens is that those _links_ are location change emitted in a _click_ event from an `onclick` attribute.

```html
<aside onclick="document.location.href='https://www.solidarite-numerique.fr/tutoriels/utiliser-le-decodex-pour-verifier-les-fausses-informations-ou-fake-news/?thematique=sinformer';" class="sinformer"></aside>
```

Not sure if Wombat should detect and fix this…

Firefox doesn't exhibit the same behavior if you're online as it will display the online content.

benoit74:
Fix #209 

Changes:
- add support for onxxx events, with same logic as wabac (https://github.com/webrecorder/wabac.js/blob/main/src/rewrite/html.js#L143-L145)
- add few test cases on test website for onclick HTML events
- fix a test which was failing due to a previously introduced trick

### DIFF
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -67,6 +67,12 @@ class HtmlRewriter(HTMLParser):
             self.url_rewriter, base_href=self.base_href, rewrite_all_url=False
         )
         self.css_rewriter = CssRewriter(self.url_rewriter, self.base_href)
+        self.js_rewriter = JsRewriter(
+            url_rewriter=self.url_rewriter,
+            base_href=self.base_href,
+            extra_rules=get_ds_rules(self.url_rewriter.article_url.value),
+            notify_js_module=self.notify_js_module,
+        )
 
         self.feed(content)
         self.close()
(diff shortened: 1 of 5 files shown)


src/warc2zim/content_rewriting/html.py
```diff
@@ -206,6 +207,8 @@ def process_attr(
             return (attr_name, ", ".join(new_value_list))
         if attr_name == "style":
             return (attr_name, self.css_rewriter.rewrite_inline(attr_value))
+        if attr_name.startswith("on") and attr_name[2] != "-":
```
rgaudin, line 210: Won't that crash if atr name is just "on"?

rgaudin (approved):
LGTM 

Do we have a ticket for improving client side dynamic rewrite?
I know it's overkill at this point but there are ways like storing a sorted list of all entries paths that might be discussed in the future.

src/warc2zim/content_rewriting/html.py
benoit74: Indeed. Added testcases and adapted code with `if attr_name.startswith("on") and not attr_name.startswith("on-")` (probably even easier to read anyway)

benoit74:
> Do we have a ticket for improving client side dynamic rewrite?

I just opened https://github.com/openzim/warc2zim/issues/276 ; I'm not sure it is a good idea to implement this, at least there are some drawbacks to discuss / find countermeasures.

Issue #271, opened 2024-05-21
Add support for real fuzzy matching


benoit74:
This issue is a placeholder for what looks like a potential enhancements warc2zim might need to implement as some point in the future (typically in a 3.x version). It is meant to summarize the current understanding of the situation and to document issues really encountered in the wild.

## Current situation (as of warc2zim 2.0)

Currently, when statically and dynamically rewriting a URL (including when computing the ZIM path of a given WARC record) the scraper applies what is called fuzzy rules. This fuzzy rule term comes from wabac vocabulary.

However, currently the scraper does not really fuzzy match, but rather only simplify/transform the ZIM path (URL has already been transformed into a ZIM path when fuzzy rules are applied).

Sample rule (in Python):
```
  {
    "pattern": r".*googlevideo.com/(videoplayback(?=\?)).*[?&](id=[^&]+).*",
    "replace": r"youtube.fuzzy.replayweb.page/\1?\2",
  }
```

The scraper checks all fuzzy rules in the list configured. The first rule with a `pattern` matching the ZIM path currently being rewritten is used, and the ZIM path is replaced by the `replace` expression.

In static URL rewriting (Python), the rewritten ZIM path is then checked for existence within the list of expected ZIM entries, and if missing URL is not rewritten.

In dynamic URL rewriting (Javascript), we do not have the list of existing ZIM entries and hence always apply the rewriting.

## Limitations

The problem with this approach appears when we have situations like graceful loading of image resolutions. E.g. the server has 4 image resolution available: `image_thumb.png`, `image_low.png`, `image_med.png`, `image_high.png` (this is a simplification / illustration of what is present on Youtube and Vimeo video placeholders, as well as article images on ir.voanews.com). The HTML document contains the `image_thumb.png`. Once DOM is loaded (typically), JS fires and replace the image `src` attribute with the proper image depending on viewport resolution (e.g. `image_low.png` on mobiles, `image_med.png` on tablets, `image_high.png` on desktop).

Since we usually scrape with a single device in Browsertrix crawler, the WARC contains only two images (usually `image_thumb.png` and `image_low.png` since we use a mobile device simulation).

But the problem is that when the user read the ZIM file on a desktop, the JS detects a big viewport and request the `image_high.png`.

Making this work is possible only with a hack: introduce two fuzzy rules:
- on matching exactly (and only) the image always requested (`image_thumb.png` in our example) so that scraper does not really rewrites this URL
[...]

PR #274, opened 2024-05-22
Fix youtube thumbnails

### LINKED ISSUE #262: Youtube video not displaying placeholder image and not playing
At https://dev.library.kiwix.org/viewer#mes-quartiers-chinois_fr_all_2024-05/mesquartierschinois.wordpress.com/

video poster is not displayed (black video) and playing video foes not work

benoit74:
Fix #262 

Changes:
- new fuzzy rule for youtube thumbnails
  - pattern has been slightly enhanced compared to the one suggested in issue to support any file extension

### DIFF
--- a/rules/rules.json
+++ b/rules/rules.json
@@ -9,6 +9,10 @@
             "replace": "youtube.fuzzy.replayweb.page/\\1\\2"
         },
         {
+            "pattern": "i\\.ytimg\\.com\\/vi\\/(.*?)\\/.*?\\.(\\w*?)(?:\\?|$)",
+            "replace": "i.ytimg.com.fuzzy.replayweb.page/vi/\\1/thumbnail.\\2"
+        },
+        {
             "pattern": "([^?]+)\\?[\\d]+$",
             "replace": "\\1"
         },
--- a/tests/test_fuzzy_rules.py
+++ b/tests/test_fuzzy_rules.py
@@ -90,6 +90,39 @@ def test_fuzzyrules_google_video_infos(google_video_info_case):
 @pytest.fixture(
     params=[
         ContentForTests(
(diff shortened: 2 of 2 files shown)


benoit74:
@mgautierfr I assigned you the review because you are probably more experienced than @rgaudin on these fuzzy rules and I would like you to check if I'm right in what I'm saying on the issue regarding the fact that there is no better way yet to be sure to always use the higher res image present in the WARC / ZIM.

@rgaudin your comments are of course still welcomed ^^

benoit74:
Just rebased the commit so that it is ready to be merged, commit content did not changed.

mgautierfr:
> I assigned you the review because you are probably more experienced than @rgaudin on these fuzzy rules and I would like you to check if I'm right in what I'm saying on the issue regarding the fact that there is no better way yet to be sure to always use the higher res image present in the WARC / ZIM.

I think there is two problems here:
- What the scrapper does (and so what is present in the warc)
- What we do

If the scrapper visits only one thumbnail (and so there is only one thumbnail per video in the warc), it seems we don't have more to do than save the thumbnail using a generic name and always rewrite url to this generic name. This is what the PR does

If the scrapper visits several thumbnails per video, we may want to be smarted on zim file creation. Instead of a "dummy rewrite" looping on all entries and processing each entry at once, we could have advance domain specific rewrite which search the best thumbnail for each video and save only this one (using a generic name here too).

I don't know if we need this advance rewrite (and I don't know how it would integrate in our current architecture neither). But in any case, we would need a generic (and unique per video) name for the thumbnail and this PR is doing that.

So in itself this PR seems good. Advance rewrite (if done) should go in another PR.

benoit74:
Thank you.

As highlighted in the linked issue, the crawler fetches multiple thumbnails per video for some videos (not all, don't know why).

Advanced rewrite is the topic of https://github.com/openzim/warc2zim/issues/271

PR #278, opened 2024-05-24
Avoid and detect redirection loops

### LINKED ISSUE #246: Warc2zim hanged forever
Task: https://farm.openzim.org/pipeline/b7a1a162-671c-4cb5-bda9-8b4b917efbc2/debug

This is most probably a recent regression in Warc2zim 2

At 2024-05-11 08:07:42, warc2zim started

At 2024-05-11 08:08:36, it had collected the metadata

Then nothing more in the log for almost 48h, so I cancelled the task at 2024-05-13 06:27

benoit74:
Fix #246 

Changes:
- do not consider redirections to same ZimPath as something to be processed (e.g. redirecting from http://www.kiwix.org to https://www.kiwix.org is "de-facto" handled in the ZIM since there is only one ZimPath for both URLs)
- still detect redirection to same ZimPath in the while loop to avoid dead loops running forever
- more properly handle the case of a nested redirection loop ( A -> B -> A ) by displaying a clear warning and not adding anything inside the ZIM
- add more logs for easier analysis with debugging the code

### DIFF
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -391,7 +391,10 @@ class Converter:
                         redirection_zim_path = normalize(
                             HttpUrl(urljoin(url, redirect_location))
                         )
-                        self.redirections[zim_path] = redirection_zim_path
+                        # Redirection to same ZIM path have to be ignored (occurs for
+                        # instance when redirecting from http to https)
+                        if zim_path != redirection_zim_path:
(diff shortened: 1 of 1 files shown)


benoit74:
Just rebased on top of main instead of warc2zim2

rgaudin (approved):
LGTM

PR #279, opened 2024-05-24
Do not rewrite URLs composed of just a fragment

### LINKED ISSUE #277: Do not rewrite href containing only a fragment
On hra-news.org, we have some anchor tag in HTML document with an href which is just an anchor like `<a href="#somevalue">`

image

This is rewritten as `<a href="./#tab2-3">` which unfortunately breaks the JS on hra-news.org (even if it is probably a more valid URL)

I suggest the scraper should detect such URL values starting with a `#` and keep these URLs as-is. 

Rewritten URL is anyway probably a bit wrong because it means nothing to have a path just like `./`

benoit74:
Fix #277 

Changes:
- Do not statically rewrite URLs composed of just a fragment (dynamic rewrite in JS is already OK)
- Add a test case to the test website with this situation and few others around the use of a folder as last portion of the URL instead of a filename

### DIFF
--- a/src/warc2zim/url_rewriting.py
+++ b/src/warc2zim/url_rewriting.py
@@ -275,6 +275,10 @@ class ArticleUrlRewriter:
         try:
             item_url = item_url.strip()
 
+            # Make case of standalone fragments more straightforward
+            if item_url.startswith("#"):
+                return item_url
+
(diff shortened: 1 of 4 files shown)


benoit74:
Just rebased on top of main instead of warc2zim2

rgaudin (approved):

PR #282, opened 2024-05-27
Properly detect nested redirection loops

### LINKED ISSUE #281: Redirection loops are still conducting to dead loops
Redirections loops like A -> B -> A are now correctly detected by changes in #https://github.com/openzim/warc2zim/pull/278

Some redirections seems to still cause dead loops.

Analysis still has to be finished, but it is sure that current logic fails to detected redirections loops like A -> B -> C -> B (it only stops if we go back to original item)

benoit74:
Fix #281 

## Changes

- Properly detect nested redirection loops like A -> B -> C -> B
- Enhance logging around redirection handling
- Add test-website test-case
- Embed test-website minimal warc (just redirection stuff) in automated Python test suite


rgaudin (approved):
LGTM ; I trust the tests 🙃

Issue #284, opened 2024-05-28
Share FuzzyRules test data between JS and Python


benoit74:
Since fuzzy rules are supposed to perform the same in JS and Python, we should define the test sets in a shared data-driven (JSON?) file so that the tests are performed the same in both languages.

PR #285, opened 2024-05-28
Fix fuzzy rule for Youtube thumbnails in JS

benoit74:
While fuzzy rule is working well in Python, trailing characters after the `?` from the querystring are not removed in Javascript causing the fuzzy rewriting to be incorrect.

E.g. `'i.ytimg.com/vi/-KpLmsAR23I/maxresdefault.jpg?sqp=-oaymwEmCIAKENAF8quKqQMa8AEB-AH-CYAC0AWKAgwIABABGHIgTyg-MA8=&rs=AOn4CLDr-FmDmP3aCsD84l48ygBmkwHg-g` is transformed into `i.ytimg.com.fuzzy.replayweb.page/vi/-KpLmsAR23I/thumbnail.jpgsqp=-oaymwEmCIAKENAF8quKqQMa8AEB-AH-CYAC0AWKAgwIABABGHIgTyg-MA8=&rs=AOn4CLDr-FmDmP3aCsD84l48ygBmkwHg-g` instead of `i.ytimg.com.fuzzy.replayweb.page/vi/-KpLmsAR23I/thumbnail.jpg`

This PR fixes the situation by updating the fuzzy rules and adding a minimal test set in Javascript.

Long term solution to test all fuzzy rules in JS is described in #284

### DIFF
--- a/javascript/test/fuzzyRules.js
+++ b/javascript/test/fuzzyRules.js
@@ -1,0 +1,35 @@
+import test from 'ava';
+
+import { applyFuzzyRules } from '../src/wombatSetup.js';
+
+test('i.ytimg.com_1', (t) => {
+  t.is(
+    applyFuzzyRules(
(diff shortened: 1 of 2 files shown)


rgaudin:
I dont think I can be of much help here but the problem description is too vague IMO.

> trailing characters are not removed in Javascript

benoit74:
Good point, I added an example in first comment.

PR #287, opened 2024-05-30
JS URL rewriting: take into account that url might be an object

benoit74:
In some occasions, url passed to JS rewriting function is an object instead of a string, causing problems like https://github.com/openzim/zim-requests/issues/1011#issuecomment-2135505415

### DIFF
--- a/javascript/src/wombatSetup.js
+++ b/javascript/src/wombatSetup.js
@@ -72,6 +72,9 @@ export function urlRewriteFunction(
 ) {
   if (!url) return url;
 
+  // Transform URL which might be an object (detected on Chromium browsers at least)
+  url = String(url);
+
   // Special stuff which is not really a URI but exists in the wild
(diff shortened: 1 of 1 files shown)


rgaudin (commented):
It really looks like you're hiding stuff under a rug here 😅 Why are you receiving a different type than the expected one? What does it look like? Why would it work once converted to String? Or maybe it doesnt but the rest of the chain doesnt complain as its a String…

benoit74:
I honestly have no idea. This comes at least from wombat, but I strongly suspect it comes from the Youtube player itself which is setting the `src` of the image to an object instead of a `String` for some reason. Tests have shown it is indeed working once converted to a String, and value is really correct, so this value is "mostly" a string (otherwise the image won't load anyway) but has not yet been converted.

I'm not totally satisfied either, but since this is also specific to one kind of reader, I propose to merge as-is for now, and we will diagnose once we are bitten again (which I suppose won't happen, but I'm well aware I might be wrong)

PR #292, opened 2024-05-31
Fix handling of <script> type in HTML documents

### LINKED ISSUE #283: BCD tables in developer.mozilla.org are broken again
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section vs https://dev.library.kiwix.org/viewer#developer.mozilla.org_en_all_2024-05/developer.mozilla.org/en-US/docs/Web/HTML/Element/section

benoit74:
Fix #283 

Changes:
- Add support for application/javascript and text/javascript
- Fix support for application/json
- Add support for all unknown script types: do not rewrite at all (instead of assuming it is JS-classic)
- Move javascript test to proper isolated test method + add more tests

### DIFF
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -88,9 +88,14 @@ class HtmlRewriter(HTMLParser):
         self.html_rewrite_context = tag  # default value if not overriden later on
         if tag == "script":
             script_type = self.extract_attr(attrs, "type")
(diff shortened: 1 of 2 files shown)


benoit74:
@rgaudin up? (no hurry but you've reviewed more recent and less "important" stuff on zimit, so I wonder if this PR might have slip)

rgaudin (approved):
Looked at it at that time but forgot to hit the button 😅

Issue #293, opened 2024-05-31
Zimit2: HTML demos on MDN pages are not working


benoit74:
![image](https://github.com/openzim/warc2zim/assets/7102089/eb77facc-04c8-4fc7-a2f1-db956513b824)

See https://dev.library.kiwix.org/content/developer.mozilla.org_en_all_2024-05/interactive-examples.mdn.mozilla.net/pages/tabbed/section.html for isolated test case.

This worked well in zimit1 and is working well in replayweb.page

URL and other arguments passed by wombat to our URL rewriting function are unusual:

```
urlRewriten:
	- current_url: http://135.181.181.97:8888/content/devmdn-bcd_2024-05/interactive-examples.mdn.mozilla.net/pages/tabbed/section.html
	- orig_host: interactive-examples.mdn.mozilla.net
	- orig_scheme: https
	- orig_url: https://interactive-examples.mdn.mozilla.net/pages/tabbed/section.html
	- prefix: http://135.181.181.97:8888/content/devmdn-bcd_2024-05/
	- url: http://mp_/blob:57e8dbed-cfaf-4e0b-a259-a23170aaf29c/https://interactive-examples.mdn.mozilla.net/pages/tabbed/section.html
	- useRel: false
	- mod: if_
	- doc: undefined
	- finalUrl: http://135.181.181.97:8888/content/devmdn-bcd_2024-05/mp_/blob%3A57e8dbed-cfaf-4e0b-a259-a23170aaf29c/https%3A//interactive-examples.mdn.mozilla.net/pages/tabbed/section.html
```

mod = `if_` and url with `mp_` seems pretty special

benoit74:
Postponing to 2.1, not an easy feat to be solved.

Jaifroid:
It looks like this **might** be related to the encoding of the querystring? In PWA I see this. Note that the app tried to call the contents with an encoded querystring, as we had decided, but the contents are actually stored in the ZIM with a plain question mark and other separators unencoded.

![image](https://github.com/openzim/warc2zim/assets/4304337/d81c3c3d-ee9c-4a4f-a2c8-b27374dd07a9)

Jaifroid:
Apologies, that was a red herring. It's just that that part of the console dev tools shows the requests in this way (encoded). It's the same for video that DOES work, so it's not that.

These code demo panels seem to work fine when it's a case of showing how JavaScript functions work, but they don't work when the content is HTML. Which is a bit strange.

[...]

Issue #239, opened 2024-04-29
Zimit2: confirm last wombat setup arguments


benoit74:
In Wombat setup code in warc2zim2, we have few arguments which are undocumented / weird.

```
    enable_auto_fetch: true,
    convert_post_to_get: true,
    target_frame: '___wb_replay_top_frame',
    isSW: true,
```

We need to confirm whether they are correct and why

Jaifroid:
I can't help with the other values, but `target_frame: '___wb_replay_top_frame'` seems like a hangover from classic Zimit, where there is an actual iframe with that identifier that contains the rendered document, and which Wombat has to be injected into (I know because I set that up more or less "manually" in the classic Zimit support in KJS). That *should* no longer be relevant (but I've no idea if warc2zim2 works without it...).

PR #294, opened 2024-05-31
Do not indicate to wombat that we are running inside a SW since we are not

benoit74:
This is a draft PR because while it allow to solve the problem in #293, it breaks youtube videos (and probably other things)

I don't really get yet why we used to indicate to wombat that we are running inside a Service Worker since we are not.

Git history is not very clear, and this was one of the undocumented option indicated to be solved in https://github.com/openzim/warc2zim/issues/239 ; @mgautierfr do you have any explanation to share?

I tested this with all test cases from the test website and they seems to all responded positively except the Youtube video.

### DIFF
--- a/javascript/src/wombatSetup.js
+++ b/javascript/src/wombatSetup.js
@@ -288,7 +288,7 @@ export function getWombatInfo(
     enable_auto_fetch: true,
     convert_post_to_get: true,
     target_frame: '___wb_replay_top_frame',
-    isSW: true,
+    isSW: false,
   };
 }
 


mgautierfr:
> do you have any explanation to share?

Not really. Some mix of copy/paste and "it was working".
BTW, I don't know what is the real impact on wombat when we set (or not) this config option.

Jaifroid:
Speculation: The Replay Worker could be run either as a Service Worker or as a Web Worker. Is it possible that some URLs would be rewritten differently if they were intended to be caught by a Service Worker as opposed to being "sent" to a regular Worker?

benoit74:
Superseeded by #334

PR #299, opened 2024-06-04
Drop integrity attribute in HTML `<script>` and `<link>` tags

### LINKED ISSUE #298: Drop `integrity` attribute in `<script>` HTML tags
See https://github.com/openzim/zim-requests/issues/1025

We cannot click on the navigation tree on the left (at least, it does not expand / collapse sections properly).

Broken ZIM is available at https://dev.library.kiwix.org/viewer#dart.dev_en_all_2024-06

benoit74:
Fix #298 

Since the JS and CSS scripts are rewritten, integrity checks are failing on purpose. Since we do not really need these integrity checks, we just drop the attribute when rewritting the HTML.

Edit:
We do it for both <script> and <link> tags, because both are allowed to use this attribute

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Drop `integrity` attribute in HTML `<script>` and `<link>` tags (#298)
+
 ## [2.0.0] - 2024-06-04
 
 ### Added
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -128,6 +128,7 @@ class HtmlRewriter(HTMLParser):
             self.transform_attrs(
                 attrs,
                 self.url_rewriter_existing if tag == "a" else self.url_rewriter_all,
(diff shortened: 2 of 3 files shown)


rgaudin:
I have a vague memory of a lib not working as expected without its integrity attr. Can you investigate? Might well be an incorrect memory…

rgaudin (approved):
LGTM ; should probably be applied to `<link />` as well, as noted in the issue.

I don't see how removing it could be an issue (except security wise) so please disregard my previous comment.

benoit74:
I failed to find a mention of a lib not working as expected without its integrity attr in Google searches. 

Not saying that you have incorrect memories, but without more information it will hard to consider it as mandatory.

Especially since technically this information is not passed to the JS, but it is true that the JS could check the DOM to check how it was loaded and wether an `integrity` attribute was still present. 

I suggest to keep this PR as-is, and adjust once we are sure that the scenario exist in the wild. Recomputing the shasum is feasible but still a bit of unnecessary work if it does not really exist in the wild.

I've also updated the PR to consider `<link>` tags as well since it looks like they are using this integrity attribute as well (see https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity).

benoit74:
Our comments have crossed, but at least we are on the same page ^^

PR #302, opened 2024-06-10
Use same automatic encoding detection for all contents

### LINKED ISSUE #301: Automatic detection of encoding not used for JS, JSON (and CSS) files
Currently, we have a custom detection of file encoding, but unfortunately this is used only for HTML content.

This is not used for JS and JSON content, making some of these rewrites fail.

For CSS, we let tinycss use its own logic to detect encoding.

We should use the same automatic detection of encoding for all files (including CSS for rationality).

benoit74:
Fix #301 

Changes:
- revisit decode of WARC content from bytes to string
  - only HTML content was benefiting from automatic encoding detection. 
  - CSS used custom tinycss detection method
  - now, all content type (HTML, JS, CSS and JSON) use the same automatic encoding detection.
- add test cases around handling of encoding by the generic rewriter

### DIFF
--- a/src/warc2zim/content_rewriting/generic.py
+++ b/src/warc2zim/content_rewriting/generic.py
@@ -186,7 +186,7 @@ class Rewriter:
 
     @no_title
     def rewrite_css(self) -> str | bytes:
-        return CssRewriter(self.url_rewriter, base_href=None).rewrite(self.content)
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
That's a lot better 👍

PR #303, opened 2024-06-10
Set correct charset in HTML documents

### LINKED ISSUE #253: Charset declared in HTML documents are not rewritten
While we always rewrite all documents in UTF-8 (as recommended by the ZIM specification), we do not update accordingly the charset declared in the HTML document headers

When present, both the `Content-Type` and the `charset` meta should probably be updated to always indicated UTF-8. Most browsers do not care much about these values and have implemented nice fallbacks, so it is not an urgent thing to fix, but probably still important to produce valid HTML documents inside the ZIM.

E.g.

```
	<meta http-equiv="Content-Type" content="text/html; charset=UTF-32" />
	<meta charset="UTF-32" />
```

Should be fixed to

```
	<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
	<meta charset="UTF-8" />
```

benoit74:
Fix #253

### DIFF
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -129,6 +129,16 @@ class HtmlRewriter(HTMLParser):
                 attrs,
                 self.url_rewriter_existing if tag == "a" else self.url_rewriter_all,
                 ["integrity"] if tag == "script" or "link" else [],
+                rewrite_content_attr=(
+                    tag == "meta"
+                    and any(
+                        attr_name.lower() == "http-equiv"
+                        and attr_value
+                        and attr_value.lower() == "content-type"
+                        for (attr_name, attr_value) in attrs
+                    )
+                ),
+                rewrite_charset_attr=(tag == "meta"),
             )
         )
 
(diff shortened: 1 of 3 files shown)


src/warc2zim/content_rewriting/html.py
```diff
@@ -215,6 +228,12 @@ def process_attr(
             return (attr_name, self.css_rewriter.rewrite_inline(attr_value))
         if attr_name.startswith("on") and not attr_name.startswith("on-"):
             return (attr_name, self.js_rewriter.rewrite(attr_value))
+        if attr_name == "content" and rewrite_content_attr:
```
rgaudin, line 231: I find this very fragile and error prone: having the condition split at two different parts of the code.

rgaudin (approved):
See my comment ; I think it deserves a comment for easier maintenance.

src/warc2zim/content_rewriting/html.py
benoit74: Just right. I found it pretty nasty to write as well. For now I'll just add a comment, but I've opened https://github.com/openzim/warc2zim/issues/305 to refactor this part of the codebase.

PR #304, opened 2024-06-11
Add support for multiple ZIM langs + validate Language metadata

### LINKED ISSUE #300: Add support for multiple languages in `--lang` / ZIM metadata
Currently, it is not possible to specify multiple languages in ZIM metadata.

We need support for the `--lang` parameter to be a list of ~semi-colon `;`~ comma `,` separated language codes.

benoit74:
Fix #300 

Changes:
- isolate method to parse language received either from input (CLI) args or automatically retrieved from HTML metadata
- add unit testing on this method
- add support for a list of languages, permissive to spaces which will be trimmed and preserving list of languages order
- add validation of language metadata, including early validation if the lang comes from input (CLI) args
- add a real fallback to `eng` as indicated in usage, wasn't done when no lang was provided at CLI and no lang was found in WARC records

### DIFF
--- a/src/warc2zim/language.py
+++ b/src/warc2zim/language.py
@@ -4,10 +4,31 @@ from warc2zim.constants import logger
 
 
 def parse_language(input_lang: str) -> str:
-    """Transform the input language into a valid ZIM Language Metadata"""
-    try:
-        lang_data = get_language_details(input_lang)
-        return lang_data["iso-639-3"]
-    except Exception:
-        logger.error(f"Invalid language setting `{input_lang}`. Using `eng`.")
-        return "eng"
+    """Transform the input language into a valid ZIM Language Metadata
+
+    Support many ways to describe the language (all ISO 639 + English label)
+    Supports a comma-separated list of languages as input.
+    Deduplicates languages.
+    Ignore whitespaces.
+    Preserve language ordering (since it conveys meaning in ZIM metadata).
+    """
+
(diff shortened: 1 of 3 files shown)


src/warc2zim/language.py
```diff
@@ -0,0 +1,28 @@
...
+
+    for lang in [lang.strip() for lang in input_lang.split(",")]:
+        try:
+            lang_data = get_language_details(lang)
+            parsed_lang = lang_data["iso-639-3"]
+            if parsed_lang not in langs:
+                langs.append(parsed_lang)
+        except Exception:
+            logger.error(f"Invalid language setting `{lang}`. Using `eng`.")
+            return "eng"
```
rgaudin, line 26: Looks like a `fra,unkown` would return `eng` while I would expect `fra`.
Shouldn't we log and skip un-parsable ones instead?

rgaudin (changes requested):

src/warc2zim/language.py
benoit74: You're right, this is silly.
Let's log and skip un-parsable ones.

benoit74:
I've also changed log level to WARNING since these problems are recovered, and I added a WARNING log when the general fallback is applied (when no language has been specified at input - CLI - and no language has been found in main entry)

rgaudin (approved):
👍

PR #310, opened 2024-06-13
Upgrade deps + fix rewrite mode

benoit74:
Changes:
- upgrade dependencies before release
- fix return value of `get_rewrite_mode` when mode is retrieved based on new `WARC-Resource-Type` header
  - including a test checking that resources have been really rewritten, to catch this bug
- javascript and JSON resources can also have `xhr` as `WARC-Resource-Type`


benoit74:
@rgaudin just a quick look is probably enough, not much in this PR

rgaudin (approved):
LGTM

PR #314, opened 2024-06-14
Fix detection of encoding again

### LINKED ISSUE #312: Automated encoding detection is still not working properly
Before https://github.com/openzim/warc2zim/pull/302, all JS / JSON (and CSS) documents were supposed to be encoded in UTF-8.

This was unreliable because some sites are not using "UTF-8" for encoding these documents.

The PR hence modified code to use the "automated detection" already in place for HTML document.

With this automated detection algorithm, we try in order:
- if available, try to use encoding received from HTTP headers
- if available, try to use encoding found in first bytes of the document, based on a regex (hopping first bytes are only using the ascii table)
- try to use `chardet` most probably encoding
- if available, fallback to encoding received from HTTP headers ignoring all bad chars

Unfortunately, it looks like this automatic detection is not that reliable. This is especially visible for JS because when no encoding is received from HTTP header, we do not have encoding specified in first bytes of the document either, and so it means we are back to simply relying on the fact that `chardet` most probable encoding is going to work.

While it gave good results on the files we tested, it seems that `chardet` is also very poorly performing on other situations.

E.g. it is failing to properly decode "https://www.cloudflare.com/vendor/onetrust/scripttemplates/202308.2.0/otBannerSdk.js" (which is just UTF-8 but detected as Windows-1252 by chardet after a very long heuristic - about 3 secs on my Linux server).

Given all these problems, it is now clear that we first need to assemble a test set of files that are now known to be difficult to decode based on our experience and gather strings in those files which we know how they should be decoded.

Then based on this test set we will be able to decide whether an automated approach still seems feasible (and which one) or if it is just impossible and the only most reasonable compromise is to allow to specify encoding to use when unknown at the CLI (with potentially multiple encoding needed per conversion, so potentially needing pattern matching rules ...).

**Nota:** my hopes for an automated solution are decreasing ; while researching a bit the web I discovered that even "big" libraries like httpx are struggling on the matter. It looks like they started with `chardet`, then switched to fully manual heurisitic (https://github.com/encode/httpx/pull/1269) and are now using `charset-normalizer` (https://github.com/encode/httpx/pull/1791). And "spoiler", `charset-normalizer` is not properly decoding the content at https://www.marxists.org/espanol/menu.js (which is one of our test cases). 

We should also need to keep in mind that bad characters exists "for real" on some document on the web (see https://github.com/openzim/warc2zim/issues/221 where we have a document from https://www.solidarite-numerique.fr/tutoriels/comprendre-les-cookies/ which is mostly only UTF-8 chars - accent works as expected, ... - but contains a bad character which is impossible to decode in UTF-8) which make this decoding task even harder.

benoit74:
Fix #312 

Changes

- first try to find charset in document first 1024 bytes with the ascii *or UTF-16 or UTF-32* encoding and regex ; if found, decode with this encoding in 'replace' mode (always going to work, do not care if encoding specified was wrong, not our fault)
  - NOTA: `or UTF-16 or UTF-32` is a slight change compared to issue proposal because with UTF-16 and UTF-32 first bytes cannot be decoded to ascii properly (they always use at least 16 bytes or 32 bytes even for ascii characters)
- if no charset in found in document first 1024 bytes, search for charset in HTTP Content-Type header ; if found, decode with this encoding in 'replace' mode (always going to work, do not care if encoding specified was wrong, not our fault)
- if no charset found in document or HTTP header, we will have to guess a little bit (situation is quite common for JS and CSS); try UTF-8 in 'strict' mode ; if it fails, try ISO-8859-1 in 'strict' mode ; if it fails, stop the scraper (we cannot guess encoding when not specified in document or HTTP header) ; this list of charset to try should even probably be exposed as a scraper parameter --guessing-charsets so that it could be possible when needed to tweak this list when needed


benoit74:
CodeFactor is complaining about files which should be ignored since they are outside our scope, these are sample files from online websites.

rgaudin:
Can you add exceptions then? Either inline, via a config file or via Codefactor UI

rgaudin (approved):
Can we rename `guessed_charsets` with `charsets_to_try` ?

Good luck with Codefactor 😀

benoit74:
Codefactor issues its configuration from `main` branch, so you have to merge first to `main` before you can review the PR ...

rgaudin:
Go ahead

kelson42:
@benoit74 Bravo... and bon courage!

benoit74:
> Go ahead

I already pushed only the configuration to main branch. My comment was more for "the posterity". I still have to rewrite this branch to change arg name + simplify commits.

> Bravo... and bon courage!

We will finish by nailing this down ^^

PR #319, opened 2024-06-17
Add option to ignore charsets found automatically

### LINKED ISSUE #318: Add options to ignore charsets found automatically
In order to cope with lying websites, it would be useful to have two options:
- one to ignore the charset found automatically in content first bytes (typically HTML documents)
- one to ignore the charset found automatically in content HTTP headers `Content-Type`

While quite "rough" (this change the scraper behavior for the whole website), this will allow websites like http://www.bouquineux.com to still be scrapped properly. This website seems to systematically indicate that HTML content is encode with `iso-8859-1` while this fails to decode any accentuated characters, and it really looks like this has been encoded with UTF-8. See e.g. http://www.bouquineux.com/?ebooks=164&Carco for improperly encoded characters. Without these new settings, content is improperly decoded, and since accentuated characters are present even in URL (where they should be URL encoded but ...), the ZIM is simply not usable at all.

benoit74:
Fix #318 

Changes:
- add the two options mentioned in issue and use them to ignore automatically detected charsets
- add missing CHANGELOG entries from previous PRs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `--ignore-content-header-charsets` option to disable automatic retrieval of content charsets from content first bytes (#318)
+- Add `--ignore-http-header-charsets` option to disable automatic retrieval of content charsets from content HTTP `Content-Type` headers (#318)
+
 ## [2.0.1] - 2024-06-13
 
 ### Added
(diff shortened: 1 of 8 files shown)


benoit74:
Nota: tested successfully on www.bouquineux.com

benoit74:
Sorry, I committed and pushed to the wrong branch. I reset the PR as it was when review was requested, sorry for that.

rgaudin (approved):
Hope we won't end up with a cli option per recipe 😉

LGTM

benoit74:
> Hope we won't end up with a cli option per recipe 😉

That's a good question, it is not the last CLI option I plan to add 😉 But flying an A380 is not like flying a Piper, you need a bit more controls ^^

PR #321, opened 2024-06-17
Add option to specify content header length

### LINKED ISSUE #320: Add option to specify how characters to consider when searching charset in content header
At https://www.marxists.org/espanol/justo/suvida.htm, the charset specified in HTML header is unfortunately far away (we need 1028 bytes to find it in full, instead of the default 1024 bytes).

Currently, we arbitrarily decided to consider only the first 1024 bytes of the content to lookup for charset. While this default value makes sense as a compromise between capacity to find all charsets and performance / memory footprint, it would help a lot if we could customize this option for the rare cases like here where the content-type is specified, a bit custom (`windows-1252` here), and we don't mind to explore more bytes on all contents.

I suggest we should add an option to customize this "magic number".

benoit74:
Fix #320 

To be reviewed / merged after #319

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add `--ignore-content-header-charsets` option to disable automatic retrieval of content charsets from content first bytes (#318)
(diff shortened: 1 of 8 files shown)


rgaudin (approved):
LGTM but what I don't like here is that it's a magic option. There's no information about when this should be changed, why and how to figure out a better value.

PR #335, opened 2024-06-25
Wabac fuzzy rules - update + process

### LINKED ISSUE #216: Add new upstream wabac.js fuzzy rules and keep track of sync commit
The scraper has fuzzy rules coming from wabac.js fuzzy matcher: https://github.com/webrecorder/wabac.js/blob/main/src/fuzzymatcher.js

These rules have been updated at the source but changes have not been back-ported.

One significant thing I noticed is that some rules (e.g. https://github.com/webrecorder/wabac.js/blob/c1124c0d93e44ff0005d4df960d989c6e7295534/src/fuzzymatcher.js#L57-L65) are returning multiple URLs while we do not take this into account. This is particularly visible in the tests: https://github.com/webrecorder/wabac.js/blob/c1124c0d93e44ff0005d4df960d989c6e7295534/test/fuzzyURL.js#L26-L30.

Is it something intentional? @rgaudin @mgautierfr any memories to share around this?

Another significant change is that more websites (washingpost, twitter, ...) are now supported by Wabac.js in this file.

This is an issue for both zimit1 and zimit2, but I suggest to fix this only for zimit2. WDYT?

benoit74:
Fix #216 

Changes:
- no new rules to update, since we do not support Twitter, Washington Post, WixStatic, Facebook
- moved rules definitions to yaml (instead of json) so that we can add inline comment ... sic ... and we do not need to escape the string values
- added information about where these rules comes from, a bit on how to update them, when it was done last time
- properly install dependencies needed to generate rules


rgaudin (approved):
LGTM

> moved rules definitions to yaml (instead of json)
> so that we can add inline comment ... sic ... 

I've seen and used fake keys for this:

```json
{
    "fuzzyRules": [
        {
            "//": "Youtuble rule related comment",
            "pattern": ".*googlevideo.com/(videoplayback(?=\\?)).*[?&](id=[^&]+).*",
            "replace": "youtube.fuzzy.replayweb.page/\\1?\\2"
        }
    ]
}
```
> and we do not need to escape the string values

It's not required. It's a convention.

That's just for the sake of the argument. YAML is fine.
Aren't we planning on using a single, shared, file with browsertrix though? JS world is obviously more JSON-oriented than YAML.

benoit74:
Good points ! 

> Aren't we planning on using a single, shared, file with browsertrix though? JS world is obviously more JSON-oriented than YAML.

Yes we are, but probably not in the coming weeks. Not sure which standard we might use then

Regarding escaping in JSON string, I agree we are not forced to do it, but it's required according to https://www.ietf.org/rfc/rfc4627.txt paragraph 2.5 (it is just a memo, so somehow we could ignore it, but still ...): 

>    The representation of strings is similar to conventions used in the C
   family of programming languages.  A string begins and ends with
   quotation marks.  All Unicode characters may be placed within the
   quotation marks except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).
>
>   Any character may be escaped.  If the character is in the Basic
   Multilingual Plane (U+0000 through U+FFFF), then it may be
   represented as a six-character sequence: a reverse solidus, followed
   by the lowercase letter u, followed by four hexadecimal digits that
   encode the character's code point.  The hexadecimal letters A though
   F can be upper or lowercase.  So, for example, a string containing
   only a single reverse solidus character may be represented as
   "\u005C".
>
>   Alternatively, there are two-character sequence escape
   representations of some popular characters.  So, for example, a
   string containing only a single reverse solidus character may be
   represented more compactly as "\\".

I will merge as-is, we can always rollback, it's not like it is a big change ^^

PR #339, opened 2024-06-25
Fix logic of rewrite mode computation for cases raised in #326

### LINKED ISSUE #326: URL rewriting is broken on absolute URLs
See https://dev.library.kiwix.org/viewer#www.mankier.com_en_all_2024-06/www.mankier.com/ where all URLs are broken, they are still absolute URLs.

Same problem on https://dev.library.kiwix.org/viewer#footballdatabase_2024-06/footballdatabase.com/ranking/europe/1

Once fixed, please update https://github.com/openzim/zim-requests/issues/171 and https://github.com/openzim/zim-requests/issues/68

benoit74:
Fix #326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Exit with cleaner message when no entries are expected in the ZIM (#336) and when main entry is not processable (#337)
 
+### Fixed
+
+- Some resources rewrite mode are still not correctly identified (#326)
+
 ## [2.0.2] - 2024-06-18
 
 ### Added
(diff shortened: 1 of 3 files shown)


benoit74:
Nota: I checked also few other task I had in mind and which have completed since 2.0.1 release to confirm there are no other known cases for now: https://farm.openzim.org/recipes/100rabbits, https://farm.openzim.org/recipes/website.test.openzim.org_en_all, https://farm.openzim.org/recipes/cloudflare.com_en_learning-center)

rgaudin (approved):
LGTM but I cannot judge the strategy here.
I think we should have a comment in the code explaining the strategy: why are we not simply relying on mimetype, etc

benoit74:
Thank you a lot for the remark, I've indeed added documentation (in `technical_architecture.md`) and opened a discussion with webrecorder to see if we can enhance the situation (things have changed a bit since the original discussions about the `WARC-Resource-Type` header): https://github.com/webrecorder/browsertrix-crawler/issues/630

PR #343, opened 2024-06-28
Refactor HTML rewriter class to make it more open to change and expressive

### LINKED ISSUE #305: Enhance maintainability and accuracy of HTML rewriting rules
Currently, when we rewrite HTML tags:
- logic is hard to read because it is dispersed between the logic handling the tag and the logic handling the attributes themselves
- logic is not totally accurate because we do not take into account the whole tree hierarchy (e.g. we rewrite meta tags no matter where they are, while we should do it only in html->head->meta) ; this is not a huge concern because usually when tag should be rewritten at only one position of the tree, then it is not valid from an HTML perspective to find it somewhere else ; however it does not help to read
- logic is not open to extension of new rules: adding new rules involves precise tweaking of various parts of the code where one would prefer to just declare a new rule with its logic somewhere (probably in a class)
- logic is not really easy to test: in order to test the rules, we need to setup the whole HTML document and the whole rewriter, including the set of known ZIM items, ... even if we do not care about them

benoit74:
Fix #305 

Idea: we should be able to define individual rewriting rules with functions and decorators, just like we do when defining endpoints in FastAPI. We have one decorator per kind of modification: drop attribute, rewrite attribute, rewrite data (and maybe others) and we implement one decorated function per kind of modifications expected.

Changes:
- new HTMLRewritingRules class with decorators `drop_attribute` (to decide to drop an HTML tag attribute), `rewrite_attribute` (to rewrite a given HTML tag attribute - name + value), `rewrite_tag` (to rewrite the whole HTML tag, potentially dropping it completely or changing tag name), `rewrite_data` (to rewrite HTML tag data)
- an instance of this class is wired in the html rewriting logic to properly process HTML code
- new tests are added to unit-test the new class behavior

See https://github.com/openzim/warc2zim/pull/353 for a show case of how this new code structure make it easier to:
- extend rewriting logic with support for new case
- test in isolation the extension, without having to deal with whole HTML tree

### DIFF
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -1,22 +1,69 @@
 import io
 from collections import namedtuple
 from collections.abc import Callable
-from functools import partial
+from dataclasses import dataclass
 from html import escape
 from html.parser import HTMLParser
+from inspect import signature
 
 from bs4 import BeautifulSoup
 
-from warc2zim.content_rewriting import UrlRewriterProto
 from warc2zim.content_rewriting.css import CssRewriter
 from warc2zim.content_rewriting.ds import get_ds_rules
 from warc2zim.content_rewriting.js import JsRewriter
 from warc2zim.content_rewriting.rx_replacer import RxRewriter
 from warc2zim.url_rewriting import ArticleUrlRewriter, ZimPath
 
-AttrsList = list[tuple[str, str | None]]
+AttrNameAndValue = tuple[str, str | None]
+AttrsList = list[AttrNameAndValue]
 
 RewritenHtml = namedtuple("RewritenHmtl", ["title", "content"])
+
+
+def get_attr_value_from(
+    attrs: AttrsList, name: str, default: str | None = None
+) -> str | None:
+    """Get one HTML attribute value if present, else return default value"""
+    for attr_name, value in attrs:
+        if attr_name == name:
+            return value
+    return default
+
+
+def format_attr(name: str, value: str | None) -> str:
+    """Format a given attribute name and value, properly escaping the value"""
+    if value is None:
+        return name
+    html_escaped_value = escape(value, quote=True)
+    return f'{name}="{html_escaped_value}"'
+
+
+def get_html_rewrite_context(tag: str, attrs: AttrsList) -> str:
+    """Get current HTML rewrite context
+
+    By default, rewrite context is the HTML tag. But in some cases (e.g. script tags) we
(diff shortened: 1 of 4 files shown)


benoit74:
@rgaudin this is far from being done, but I would like to have a first feedback on this approach 

For now only the simple "drop attribute" case is implemented, but it gives a rough idea of what it will look like if I continue on this approach.

I'm a bit surprise I had to implement the `call_func` helper function myself, I'm pretty sure it is possible without this custom code and Python stdlib code, but I failed to find how to write it properly. Idea is that I want to pass whatever argument are defined in the decorated function but not other ones.

benoit74:
Refactoring work is now completed, all HTML rewrite operations are isolated in standalone methods.

src/warc2zim/content_rewriting/html.py
```diff
@@ -1,24 +1,69 @@
...
 from warc2zim.url_rewriting import ArticleUrlRewriter, ZimPath
 
-AttrsList = list[tuple[str, str | None]]
+AttrNameAndValue = tuple[str, str | None]
+AttrsList = list[AttrNameAndValue]
 
 RewritenHtml = namedtuple("RewritenHmtl", ["title", "content"])
 
 
+def extract_attr(attrs: AttrsList, name: str, default: str | None = None) -> str | None:
```
rgaudin, line 23:
```suggestion
def get_attr_value_from(attrs: AttrsList, name: str, default: str | None = None) -> str | None:
```

src/warc2zim/content_rewriting/html.py
```diff
@@ -195,81 +215,412 @@ def handle_pi(self, data: str):
...
+        for attr in [
+            (attr_name, attr_value)
+            for (attr_name, attr_value) in attrs
+            if attr_name != "href"
+        ]
+    )
+    if values:
+        return f"<base {values}{'/>' if auto_close else '>'}"
+    else:
+        return ""  # drop whole tag
```
rgaudin, line 587: Hum, seems a bit fragile but we can live with it

rgaudin (approved):
Very nice. I think the implementation is relatively clear and easily maintainable. I have no suggestion ATM. Might come with usage over time

src/warc2zim/content_rewriting/html.py
benoit74: Yep, especially since it is heavily tested

src/warc2zim/content_rewriting/html.py
benoit74: Agreed

PR #344, opened 2024-07-02
Add debug log for items whose content is empty

benoit74:
Just a small debug log to help diagnosis


benoit74:
> Worth noting that zimcheck is unhappy with empty ZIM entries

Thank you, I added a comment to make it even more explicit

PR #348, opened 2024-07-17
Generate fuzzy rules tests

### LINKED ISSUE #284: Share FuzzyRules test data between JS and Python

benoit74:
Fix #284 

Changes:
- fuzzy rules defined in YAML must now have a `name` and a `tests` properties
- tests in Python and in Javascript are automatically generated from YAML thanks to these two properties
- one rule (vimeo_cdn_fix) needed to be fixed because tests showed that it was not working as intended in JS


rules/generate_rules.py
```diff
@@ -13,12 +13,23 @@
 
 FUZZY_RULES = yaml.safe_load(rules_src.read_text())["fuzzyRules"]
 
+for rule in FUZZY_RULES:
+    if "name" not in rule:
+        print("Fuzzy rule is missing a name, exiting")
+        sys.exit()
```
rgaudin, line 19: For the sake of it, you should exit with `1` or even more explicitly raise `SystemExit("FZ rule missing")`

rgaudin (approved):
LGTM

rules/generate_rules.py
benoit74: Agreed

Issue #332, opened 2024-06-24
ValueError: Incorrect HttpUrl scheme in value


rgaudin:
[This zimit run](https://farm.zimit.kiwix.org/pipeline/fb6ad63d-68f8-4315-80da-9c7878107e69/debug) failed rewriting an URL with an exotic scheme (`intent`).

```
ValueError: Incorrect HttpUrl scheme in value: intent://show/0vT7LJMeVDxyQ2ZamHKu08?si=_uDRD6bRR_6M5YZyISG
```

Non-`http`/`https` schemes are valid and should be ignored instead of being rewritten. Not sure if this statement is not in line with current implementation or if it's just a bug…

benoit74:
If I'm not mistaken, non-http(s) schemes are ignored only when rewriting URLs. 

Here the error happens during the initial `gather_information_from_warc` where we loop over all entries, so I strongly suspect we have a WARC record associated with this URL. Didn't knew browsertrix crawler was capable of retrieving non-http(s) resources. Glad I implemented this check to avoid pushing stupid things to the ZIM. However such entries should probably just be ignored instead of raising an error and stopping the process.

benoit74:
Previous assumption is wrong, problem lies in a redirect which is targeting these URLs with weird schemes, mostly same bug as #356 (just the error message differ)

PR #351, opened 2024-07-17
Ignore non HTTP(S) WARC records

benoit74:
Might have been a solution to #332 but not (see next comment)

NOTA : there is no test because I failed to find under which conditions we can have a non HTTP(S) WARC record + we should test for all places where we iterate on WARC records, which makes it even harder to test

### DIFF
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -407,6 +407,11 @@ class Converter:
                 continue
 
             url = get_record_url(record)
+
+            # ignore non HTTP(S) URLs (intent:// for instance, see #332)
+            if not (url.startswith("http://") or url.startswith("https://")):
+                continue
+
             zim_path = normalize(HttpUrl(url))
 
             status_code = get_status_code(record)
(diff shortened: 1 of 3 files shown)


benoit74:
This is not fixing #332 in fact, I found why we have these weird URL schemes, bug is same as #356 

I propose to still merge this PR, it is valuable, should the WARC contain records which are not http/https we still need to ignore them.

src/warc2zim/converter.py
```diff
@@ -397,7 +397,12 @@ def gather_information_from_warc(self):
             if record.rec_type not in ("response", "revisit"):
                 continue
 
-            url = get_record_url(record)
+            url = str(get_record_url(record))
```
rgaudin, line 400: `get_record_url()`  has no return type but can it ever return other than an `str`? Wouldn't it be better then to set the return type?

rgaudin (approved):
OK

src/warc2zim/converter.py
benoit74: Agreed

PR #353, opened 2024-07-18
Correctly rewrite target url in HTML meta http-equiv redirect

### LINKED ISSUE #237: Add support for redirection in `meta http-equiv`
It is possible to redirect a page with a `meta http-equiv`:

```
<meta http-equiv="refresh" content="3;url=https://www.mozilla.org" />
```

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#http-equiv

So far, Zimit2 does not rewrite these links which are hence not leading to content inside the ZIM.

These links should be considered for rewrite as well.

The result will of course depend on wether the redirect target is present inside the WARC/ZIM or not.

Which in turns depends on how the WARC has been built (i.e. did the system which built the WARC waited long enough to have the redirect target captured as well?) but this is beyond the scope of this issue / warc2zim. warc2zim must just ensure to consider these URLs for rewrite.

benoit74:
Fix #237 

Changes:
- properly rewrite redirect target url when present in `<meta>` HTML tag
- add test case in test website

Nota: this is built on-top of https://github.com/openzim/warc2zim/pull/343 and should hence be reviewed only once this is reviewed/merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- New fuzzy-rule for cheatography.com (#342), der-postillon.com (#330)
 - Properly rewrite redirect target url when present in <meta> HTML tag (#237)
 
 ### Changed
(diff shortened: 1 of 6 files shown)


rgaudin (approved):
Pretty cool! Lacks changelog entry

PR #359, opened 2024-07-22
Source more ZIM metadata from WARC files

### LINKED ISSUE #357: Use `software` from `warcinfo` record to set `Scraper` metadata
Currently, scraper sets the `Scraper` ZIM metadata based on warc2zim version + an optional scraper suffix (usually passed by zimit with zimit version + browsertrix version).

This is not the optimal solution because WARC file already contain one (usually) `warcinfo` record which contains the browsertrix version in a `software` property.

We should prefer to use this information rather than fetching it indirectly in Zimit, especially since begining with https://github.com/openzim/zimit/issues/301 the WARC processed might come from a browsertrix version different than the one packaged with zimit. Even without this issue, it would allow to use any WARC harvester and still automatically have a proper `Scraper` metadata.

### LINKED ISSUE #358: Store `ContentDate` as metadata, based on WARC-Date
We should compute and store in a ZIM metadata the dates when the content has been fetched into the WARC.

This information is present in WARC records but lost once converted into a ZIM. This is a pity since it makes debugging ZIM issues even harder.

See https://github.com/openzim/overview/issues/40 for a proposition of standardization of this metadata, but we can move forward even if it is not yet standard.

benoit74:
Fix #358 
Fix #357 

Changes:
- in addition to changes needed by issues mentioned, we stop to filter warc records by type in the `iter_warc_records` generator since this is both confusing + needless ; selecting proper record type is the "using code" responsibility
and needs might vary

### DIFF
--- a/src/warc2zim/converter.py
+++ b/src/warc2zim/converter.py
@@ -33,6 +33,7 @@ import requests
 # from zimscraperlib import getLogger
 from bs4 import BeautifulSoup
 from cdxj_indexer import buffering_record_iter, iter_file_or_dir
+from dateutil import parser
 from jinja2 import Environment, PackageLoader
 from warcio import ArchiveIterator
 from zimscraperlib.constants import (
@@ -224,6 +225,11 @@ class Converter:
 
         self.scraper_suffix = args.scraper_suffix
 
+        # metadata about WARC files
+        self.warc_software = ""
+        self.warc_start = None
+        self.warc_end = None
+
         self.continue_on_error = bool(args.continue_on_error)
         self.disable_metadata_checks = bool(args.disable_metadata_checks)
         self.ignore_content_header_charsets = bool(args.ignore_content_header_charsets)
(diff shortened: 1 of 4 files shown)


src/warc2zim/converter.py
```diff
@@ -388,15 +407,46 @@ def run(self):
...
+            )
+            return
+        for warcfield in get_record_content(record).decode("UTF-8").splitlines():
+            name_and_value = warcfield.split(":")
+            if len(name_and_value) != 2:  # noqa: PLR2004
+                logger.warning(f"Unrecognized WARC field: {warcfield}")
+                continue
+            if name_and_value[0].strip().lower() != "software":
+                continue
+            self.warc_software = "," + name_and_value[1].strip()
```
rgaudin, line 427: this is ugly. you should handle the concatenation properly when setting `Scraper`

src/warc2zim/converter.py
```diff
@@ -388,15 +407,46 @@ def run(self):
...
+        if self.warc_software:
+            logger.debug("warc_software already set, ignoring warcinfo record")
+            return
+        if get_record_mime_type(record) != "application/warc-fields":
+            logger.warning(
+                f"Unsupported warcinfo record found: {get_record_mime_type(record)}"
+            )
+            return
+        for warcfield in get_record_content(record).decode("UTF-8").splitlines():
+            name_and_value = warcfield.split(":")
```
rgaudin, line 421: I believe the following is as good (worst case is value contains additional stuff) yet clearer
```suggestion
            name, value = warcfield.split(":", 1)
            if name.strip().lower() != "software":
                continue
            self.warc_software = value.strip()
```

rgaudin (changes requested):
Mostly OK.
I think that since there are ongoing discussions about storing this information and since there is no reader supporting it, we should use a safe name for now like `WARC-ContentDate` or `X-ContentDate` or something as for it not to interfere.

src/warc2zim/converter.py
benoit74: I have to admit I'm mostly ashamed, especially when I see the code setting `Scraper` metadata which is mostly as ugly.

rgaudin (approved):
LGTM

PR #365, opened 2024-07-24
Remove double slash in URLs, both in Python and JS

### LINKED ISSUE #316: Double slash in query string (and path) are not working properly
URL rewriting is not working properly when we have double `/` in query string.

E.g. if HTML document URL is `https://www.example.com/index.html?url=https://www.example.com/foo`, then everything is broken on the page in Kiwix-server due to improper URL rewriting / conflicting browser behavior / ... 

The problem is that for such an original HTML document URL, the ZIM Path is `www.example.com/index.html?url=https://www.example.com/foo` and the URL to load this in kiwix-serve is `http://yourserver/content/yourzim/www.example.com/index.html%3Furl%3Dhttps%3A//www.example.com/foo`

From a browser/server perspective, the query string is dropped (this is normal for proper operation in all readers) and we hence end-up with a double slash `//`. 

This is not a promising URL because how relative URLs are computed given this double slash is obviously not standardised. Or at least Python (used to compute relative path in scraper) and Firefox (used to display the ZIM) have different PoV on this. And we can be sure we will have very varied ways to interpret this URL and relative URLS inside the document in our various readers.

I hence recommend that when normalizing URLs into a ZIM Path, we get rid of any `//` (or `///` and so on) and simplify them to `/`. This comes with a minimal risk of collision we can most probably assume is negligible. It would also break some websites which uses the query string value in JS code and really need the proper value. We'll see how often this happens.

Nota: An alternative could have been to keep same ZIM Path and change the way we transform ZIM Path into URL by saying that a ZIM path like `foo//bar` should lead to a URL `foo/%2Fbar`. This would ensure we keep the same look to URLs displayed in browser bar (`%2F` will continue to be displayed as `/`) and we minimize the risk of collisions (we keep same number of characters). The big problem with this approach is that since the conversion can only be done when we transform the Path into a URL (we cannot change the ZIM Path, encoded chars should not be present inside the ZIM Path), then it means that all computation of relative links which are based on the ZIM Path are not working anymore and also have to be adapted ... which seems overly complex.

Nota: this bug has been observed in the wild on https://edu.gcfglobal.org/pt/criar-um-correio-eletronico/como-funciona-um-correio-eletronico/1/# where we have an iFrame to `https://support.gcfglobal.org/form/?url=https://edu.gcfglobal.org/pt/criar-um-correio-eletronico/como-funciona-um-correio-eletronico/1/`

benoit74:
Fix #316 

Changes:
- in Python and in JS, when we compute the ZIM path from a given URL, we replace all multiple occurences of `/` character by a single `/`
  - both rely on a utility function which is not public, since not meant to be reused anywhere else is the codebase
- in Python test, compare `ZimPath` value instead of comparing object, from better pytest report in case of issue (string are compared char by char and error is quickly spotted)

### DIFF
--- a/javascript/src/wombatSetup.js
+++ b/javascript/src/wombatSetup.js
@@ -59,6 +59,12 @@ export function hasAlreadyBeenRewritten(
   return false;
 }
 
+function removeSubsequentSlashes(value) {
+  // Remove all successive occurrences of a slash `/` in a given string
+  // E.g `val//ue` or `val///ue` or `val////ue` (and so on) are transformed into `value`
+  return value.replace(/\/\/+/g, '/');
+}
+
 export function urlRewriteFunction(
   current_url, // The current (real) url we are on, e.g. http://library.kiwix.org/content/myzim_yyyy-mm/www.example.com/index.html
   orig_host, // The host of the original url, e.g. www.example.com
@@ -176,7 +182,8 @@ export function urlRewriteFunction(
       : '?' + decodeURIComponent(absolute_url_parts.query).replaceAll('+', ' ');
 
   // combine all decoded parts to get the ZIM path
-  const zimPath = decoded_host + decoded_path + decoded_query;
+  const zimPath =
+    decoded_host + removeSubsequentSlashes(decoded_path + decoded_query);
 
   // apply the fuzzy rules to the ZIM path
   const fuzzifiedPath = applyFuzzyRules(zimPath);
--- a/src/warc2zim/url_rewriting.py
+++ b/src/warc2zim/url_rewriting.py
@@ -45,13 +45,7 @@ from __future__ import annotations
 
 import re
 from pathlib import PurePosixPath
-from urllib.parse import (
-    quote,
-    unquote,
-    urljoin,
-    urlsplit,
-    urlunsplit,
-)
+from urllib.parse import quote, unquote, urljoin, urlsplit, urlunsplit
 
 import idna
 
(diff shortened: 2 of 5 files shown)


javascript/src/wombatSetup.js
```diff
@@ -59,6 +59,12 @@ export function hasAlreadyBeenRewritten(
   return false;
 }
 
+function removeDoubleSlash(value) {
```
rgaudin, line 62:
```suggestion
function removeSubsequentSlashes(value) {
```

src/warc2zim/url_rewriting.py
```diff
@@ -228,11 +222,21 @@ def normalize(url: HttpUrl) -> ZimPath:
...
 
-    fuzzified_url = apply_fuzzy_rules(f"{hostname}{path}{query}")
+    fuzzified_url = apply_fuzzy_rules(
+        f"{hostname}{_remove_double_slash(path)}{_remove_double_slash(query)}"
+    )
 
     return ZimPath(fuzzified_url)
 
 
+def _remove_double_slash(value: str) -> str:
```
rgaudin, line 232:
```suggestion
def _remove_subsequent_slashes(value: str) -> str:
```

rgaudin (approved):
LGTM ; I suggested what I think is a better name

javascript/src/wombatSetup.js
benoit74: Yep

src/warc2zim/url_rewriting.py
benoit74: Yep

PR #366, opened 2024-07-24
Properly sort WARC files and log total number of WARCs found

### LINKED ISSUE #360: Properly sort WARC files
Currently, the order in which WARC files are processed is mostly random.

We should sort WARC names alphabetically so that order is deterministic AND reflect in general the order in which resources have been fetched.

This is particularly important for JS-modules where we assume that the HTML / JS using the JS module is processed before the JS module itself. If we are unlucky, this might not be the case with current random logic.

benoit74:
Fix #360 

Changes:
- compute and sort the list of WARC files at converter startup, so that it is done only once no matter the number of times `iter_warc_records` is called
- log total number of WARC files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Handle case where the redirect target is bad / unsupported (#332 and #356)
+- Fixed WARC files handling order to follow creation order (#366)
 
 ## [2.0.3] - 2024-07-24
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Properly sort WARC files and log total number of WARCs found (#366)
```
rgaudin, line 12: This is a fix, not a change IMO. 
```suggestion
- Fixed WARC files handling order to follow creation order (#366)
```

rgaudin (approved):
LGTM

CHANGELOG.md
benoit74: Agreed and fixed

PR #371, opened 2024-08-05
Compute rule signature at rule initialization to save perf

### LINKED ISSUE #370: Performance issue linked to new "extensible" HTML rewriting rules
For a very small WARC like https://github.com/openzim/warc2zim/blob/main/tests/data-special/qsl.net-encoding-alias.warc.gz, it takes more than 2 minutes to build the ZIM.

A flamegraph shows that most of the time is spent in the `rewrite_html` (expected since the HTML page in this WARC is huge) but inside this most time is spent in `inspect.signature` function. 

qsl_flame

This `signature` information should in fact be cached since it is not going to change during a warc2zim execution.

A quick change (tbc in a PR) confirms that caching this information allows to return to coherent timings (less than 20 secs, with lot of time spent parsing the HTML which is expected since HTML is huge).

qsl_flame_cached

benoit74:
Fix #370 

Changes:
- use a generic class to automatically compute function signature at rule initialization
- use cached value at "runtime"

### DIFF
--- a/src/warc2zim/content_rewriting/html.py
+++ b/src/warc2zim/content_rewriting/html.py
@@ -5,7 +5,8 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from html import escape
 from html.parser import HTMLParser
-from inspect import signature
+from inspect import Signature, signature
+from typing import Generic, TypeVar
 
 from bs4 import BeautifulSoup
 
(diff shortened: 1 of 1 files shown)


rgaudin (changes requested):
I'd like us to discuss alternatives as I find this optimization very intrusive. Haven't dug much into the issue but it looks like what lru_cache is for at first glance.

benoit74:
Replaced by https://github.com/openzim/warc2zim/pull/375

Issue #352, opened 2024-07-18
Retrieving upstream favicon to set illustration 48x48 is not smart enough


kelson42:
Scraping https://womenshistory.si.edu/ with an extensive set of good favicon/illustrations:
```
<link rel="icon" sizes="16x16" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-16x16.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-16x16.png)" />
<link rel="icon" sizes="32x32" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-32x32.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-32x32.png)" />
<link rel="icon" sizes="96x96" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-96x96.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-96x96.png)" />
<link rel="icon" sizes="192x192" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/android-chrome-192x192.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/android-chrome-192x192.png)" />
<link rel="apple-touch-icon" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/apple-icon-60x60.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/apple-icon-60x60.png)" />
<link rel="apple-touch-icon" sizes="72x72" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/apple-icon-72x72.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/apple-icon-72x72.png)" />
<link rel="apple-touch-icon" sizes="76x76" href="[https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/apple-icon-76x76.png](view-source:https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/apple-icon-76x76.png)" />
[...]

PR #355, opened 2024-07-19
Use best icon possible as ZIM illustration

benoit74:
Fix #352 

Changes in logic around finding the ZIM illustration:
- build a sorted list of potential icons to use (instead of just a "random" one)
- prefer to use best icon in this list, either from WARC (preferably, if available) or from download (if available)
- if both fails for best icon, process next icon in the list (and so forth, until we find one working)
- if all fails, fallback to default scraperlib illustration

Searching inside the WARC is still preferred to download because it is expected to be quite fast: we already know the list of expected items and the icon is usually at the very beginning of the WARC (when present).

Note that the fact that this change contradicts significantly with what has been discussed and decided in https://github.com/openzim/warc2zim/issues/202, since there is a significant chance we will download the illustration.

In #202, we said that there is probably no situation where the best icon is not already present inside the WARC and should be downloaded. This is wrong. 

This change is grounded on a real use case: https://womenshistory.si.edu/. In this use case, we have only two WARC items fetched by the crawler:
- https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon.ico (which is 16x16)
- https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-32x32.png (which is 32x32)

Both are too small for a ZIM illustration and will need to be upscaled. This is not appropriate because scraper could know that best icon possible is https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/android-chrome-192x192.png, and this file is available for download. This is now what will happen with the change in this PR.

Issue #202, opened 2024-03-04
Should we remove the favicon download fallback


benoit74:
When a favicon URL is passed at the CLI, the scraper downloads this favicon and use it as ZIM illustration.

When no favicon URL is passed at the CLI, the scraper uses the favicon he finds on the main page as ZIM illustration.

In this second case, the favicon is first searched for in WARC records.

While this favicon should be present in WARC records (and usually is), there is also a fallback to download it from the web.

@rgaudin @mgautierfr) consider that this fallback download makes no sense and should be removed (see https://github.com/openzim/warc2zim/pull/196#issuecomment-1973306252 and https://github.com/openzim/warc2zim/pull/196#issuecomment-1973567648).

I do not agree, I find it weird to allow to download the favicon with the "CLI" favicon URL but not from the "guessed" favicon URL.

It will lead to a situation where the ZIM is either not produced (if we fail the scrape when "guessed" favicon is missing from the WARC) or produced with wrong illustration (if we use default PNG when "guessed" favicon is missing from the WARC), and we finally have to set the guessed favicon URL at CLI and restart the scraper to have the proper ZIM illustration. The scraper is better like today where he tries to intelligently solve issues without manual intervention.

rgaudin:
Favicon is a resource (in browsertrix's terms) so it's not affected by scope. What scenario would have it not be present in WARC but us able to parse and download  it correctly?

benoit74:
> Favicon is a resource (in browsertrix's terms) so it's not affected by scope. What scenario would have it not be present in WARC but us able to parse and download it correctly?

Probably none, you're right

benoit74:
I.e. we should not fallback to download what is already present in the WARC or not available at all. And use the default PNG in such a situation.

benoit74:
Closed by https://github.com/openzim/warc2zim/pull/196

PR #372, opened 2024-08-07
Revisit retrieve_illustration logic to prefer best favicons

### LINKED ISSUE #369: Scraper fails when icon in main page HTML is not an HTTP URL
See https://data.fs.usda.gov/geodata/rastergateway/states-regions/states.php

```
<link href="about:blank" rel="shortcut icon">
```

Seen on https://farm.zimit.kiwix.org/pipeline/f1a1f927-a785-4f8f-b0c6-c7d69e75ed14/debug

```
  File "/usr/bin/zimit", line 8, in <module>
    sys.exit(zimit.zimit())
             ^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/zimit/zimit.py", line 585, in zimit
    run(sys.argv[1:])
  File "/app/zimit/lib/python3.12/site-packages/zimit/zimit.py", line 507, in run
    return warc2zim(warc2zim_args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/main.py", line 146, in main
    return converter.run()
           ^^^^^^^^^^^^^^^
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/converter.py", line 272, in run
    self.gather_information_from_warc()
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/converter.py", line 491, in gather_information_from_warc
    self.find_icon_and_language(record, content)
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/converter.py", line 605, in find_icon_and_language
    HttpUrl(
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/url_rewriting.py", line 74, in __init__
    HttpUrl.check_validity(value)
  File "/app/zimit/lib/python3.12/site-packages/warc2zim/url_rewriting.py", line 95, in check_validity
    raise ValueError(
ValueError: Incorrect HttpUrl scheme in value: about:blank about

```

benoit74:
This PR replaces https://github.com/openzim/warc2zim/pull/355 with another approach: do not trust HTML to get favicon sizes + if user provided a favicon, simply use it (if it is a bad favicon ... it is bad ...).

Fix https://github.com/openzim/warc2zim/issues/352
Fix #369

Changes in logic around finding the ZIM illustration:
- if user provided a favicon to use, we use it
- instead of prefering to use WARC items (or prefering to download as it was before #202), we prefer to use the most suited favicon.
- potential favicons are sourced from main HTML page.
- ll favicons are retrieved either from the WARC or downloaded to inspect their sizes.
- we use the most suited one (i.e. 48x48 or bigger if possible or the biggest one).
- we still fallback to default ZIM illustration if no favicon is found, to avoid loosing all time spent crawling the website.

Logic has been significantly revisited to not have to loop over all warc records since most potential favicons are not inside the WARC and this would waste significant time / resources on big crawls, so we prefer to store favicons on-the-fly in memory in initial `gather_information_from_warc`. The favicon is anyway used only few moments later, and we do not expect to have many huge favicons in memory.

Note that the fact that this change contradicts significantly with what has been discussed and decided in https://github.com/openzim/warc2zim/issues/202, since there is a significant chance we will download the illustration.

In https://github.com/openzim/warc2zim/issues/202, we said that there is probably no situation where the best icon is not already present inside the WARC and should be downloaded. This is wrong.

This change is grounded on a real use case: https://womenshistory.si.edu/. In this use case, we have only two WARC items fetched by the crawler:
- https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon.ico (which is 16x16)
- https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/favicon-32x32.png (which is 32x32)

Both are too small for a ZIM illustration and will need to be upscaled. This is not appropriate because scraper could know that best icon possible is https://womenshistory.si.edu//sites/default/themes/si_sawhm/favicons/android-chrome-192x192.png, and this file is available for download. This is now what will happen with the change in this PR.


benoit74:
I just added missing CHANGELOG entry in last commit

src/warc2zim/icon_finder.py
```diff
@@ -0,0 +1,84 @@
...
+    format: str | None
+
+    def __lt__(self, other):
+        return Icon._cmp_icons(self, other) < 0
+
+    def __gt__(self, other):
+        return Icon._cmp_icons(self, other) > 0
+
+    @classmethod
+    def _cmp_icons(cls, x: "Icon", y: "Icon") -> int:
```
rgaudin, line 26: I think readability would be improved with variables named `a` and `b`. `x` and `y` confuses me because those are associated with geometrical axis and thus width/height.

rgaudin: `x_dim` would be more clear with `a_min` for instannce

src/warc2zim/icon_finder.py
```diff
@@ -0,0 +1,84 @@
...
+        if x_dim > ZIM_ILLUSTRATION_SIZE:
+            if y_dim == ZIM_ILLUSTRATION_SIZE:
+                return -1  # y is perfect, x not
+            elif x_dim < y_dim:
+                return -1  # prefer biggest icon (y)
+            elif x_dim == y_dim:
+                return 0  # both icon are equivalent
+            else:
+                return 1  #  perfer biggest icon (x)
+        if y_dim >= ZIM_ILLUSTRATION_SIZE:  # and x < ZIM_ILLUSTRATION_SIZE
```
rgaudin, line 44:
```suggestion
        if y_dim >= ZIM_ILLUSTRATION_SIZE:  # x is < ZIM_ILLUSTRATION_SIZE
```
Just formatting it for its intent (reminding status) and not confuse it with a commented out condition part

rgaudin (approved):
Very good 👍

src/warc2zim/icon_finder.py
benoit74: Yep

src/warc2zim/icon_finder.py
benoit74: Yep

benoit74:
Closed by mismanipulation (typo in my terminal)

PR #373, opened 2024-08-07
Remove domain specific rules

### LINKED ISSUE #328: Confirm we really need the DS rewriting rules
See https://github.com/webrecorder/wabac.js/pull/182#issuecomment-2185726884 for the start of a discussion about the fact we might not need DS rewriting, it looks like this is already done in the crawler at crawl time.

benoit74:
Fix #328 

Domain specific rewriting rules have been introduced because they were deemed necessary.

Discussions with browsertrix recorder team has shown that these rules are not necessary because they are already applied by Browsertrix Crawler before saving the WARC.

It has the downside that if the WARC does not comes from Browsertrix Crawler, then we might miss these rules.

I consider this downside is negligible compared to the fact that testing and maintaining these rules is a hassle.


docs/functional_architecture.md
```diff
@@ -73,26 +73,12 @@ The same fuzzy rules that have been used to compute the ZIM path from a resource
...
 While this is expected to serve mostly for the dynamic case, we still applies them on both side (staticaly and dynamicaly) for coherency.
 
-## Content rewriting
+## Documents rewriten statically
 
-### DS rules
+For now warc2zim rewrites HTML, CSS and JS documents. For CSS and JS, this mainly consists in replacing URLs. For HTML, we also have more specific rewritting necessary (e.g. to handle base href or redirects with meta).
 
-DS (Domain Specific) rules patch javascript code with regular expressions matching. These rules are not related to url or path.
+Since 2.1, no specific domain specific (DS) rules are applied like it is done in wabac.JS because these rules are already applied in Browsertrix Crawler. For the same reason, JSON is not rewritten anymore (URL do not need to be rewritten in JSON because these URLs will be used by JS, intercepted by wombat and dynamically rewritten).
```
rgaudin, line 80:
```suggestion
Since 2.1, no domain specific (DS) rules are applied like it is done in wabac.JS because these rules are already applied in Browsertrix Crawler. For the same reason, JSON is not rewritten anymore (URL do not need to be rewritten in JSON because these URLs will be used by JS, intercepted by wombat and dynamically rewritten).
```

rgaudin (approved):
LGTM
I understand @kelson42's thumb up on PR description as validation of this move.

PR #375, opened 2024-08-07
Cache HTML rewriting function signature

benoit74:
This PR replaces https://github.com/openzim/warc2zim/pull/371 with another approach based on `@lru_cache`

Fix https://github.com/openzim/warc2zim/issues/370

Changes:
- use `@cache` (an `@lru_cache` without max size since we want to cache all values) to cache rewriting rule function signature


benoit74:
Compared to the approach propose in #371, performance seems almost identical, this approach is only marginally slower (19.8 secs vs 19 secs)

rgaudin (approved):
OK ; Better rely on stdlib if we can 👍

Issue #403, opened 2024-09-26
<img> tags in HTML document are not working probably due to srcset attribute


benoit74:
In banrepcultural enciclopedia ZIM (at https://dev.library.kiwix.org/#lang=&q=enciclopedia+banrepcultural), we have an issue around images which are not displaying unless we use a mobile phone (or maybe a tablet)

Sample page: https://dev.library.kiwix.org/content/banrepcultural.org_es_enciclopedia_2024-09/enciclopedia.banrepcultural.org/index.php%3Ftitle%3DDelcy_Morelos_Sandoval

HTML sample from this page:
```
<img src="images/thumb/b/b8/Avatar-mujer.jpg/300px-Avatar-mujer.jpg" decoding="async" width="300" height="341" srcset="images/b/b8/Avatar-mujer.jpg 1.5x" />
```

What is inside the ZIM related to `Avatar-mujer.jpg`:
```
enciclopedia.banrepcultural.org/images/b/b8/Avatar-mujer.jpg
enciclopedia.banrepcultural.org/images/thumb/b/b8/Avatar-mujer.jpg/105px-Avatar-mujer.jpg
```

The `enciclopedia.banrepcultural.org/images/thumb/b/b8/Avatar-mujer.jpg/105px-Avatar-mujer.jpg` seems unrelated to this page, it seems to come from https://dev.library.kiwix.org/content/banrepcultural.org_es_enciclopedia_2024-09/enciclopedia.banrepcultural.org/index.php%3Ftitle%3DArchivo%3AAvatar-mujer.jpg (the thumbnail in the table `File history`)

Two questions:
- why did the rewriter accepted to rewrite/keep `images/thumb/b/b8/Avatar-mujer.jpg/300px-Avatar-mujer.jpg` despite this path being missing in the ZIM?
- can we imagine warc2zim explores all src and srcset URLs and replace the src with the image present in the WARC/ZIM and drop the whole srcset? we already have a [rewrite rule](https://github.com/openzim/warc2zim/blob/c8e1e964479e6f3cf416b5df022c33d38c715ac0/src/warc2zim/content_rewriting/html.py#L560-L577) to properly rewrite the URLs in the srcset attribute, but I start to consider this might probably have been a mistake, only one URL has been really load in most (all?) cases and we should have fixed this differently

kelson42:
Once the version of the image chosen, it seems to me that - indeed - the whole "srcset" should be replaced by a simple "src".

benoit74:
[...]

PR #406, opened 2024-10-08
HTML document can be retrieved as `fetch`

### LINKED ISSUE #405: HTML documents can also have `resourcetype` as `fetch`
Hello, I'm running into this weird issue:
I've archived a website but most of the links in it are not being opened in the built in browser, but externally in Firefox, requiring internet to load and thus eliminating the point in archiving them. I'm going to be specific:

I ran this command in docker:
docker run  -v [mydir]:/output ghcr.io/openzim/zimit zimit --url https://www.allaboutcircuits.com/ --name nameofthezim

Now this ran for a whopping 14 hours, resulting in a nearly 70GB zim file, and I thought nice, great, amazing. However, upon opening it I saw that not only most images have not been downloaded (which is likely due to me not using autoscroll in all honesty), but most links I click lead me to an external tab in my browser, which is Firefox. Upon right clicking on the both types of links (the ones that open in the built in browser, and the ones that open in FF) I've discovered that:

The ones that open in the built in browser have the option to Open Link In New Tab, while the the ones that open in FF have the option to Open Link In Web Browser.

Needless to say, this is a very severe issue, one could say a fatal one. I don't know what these 70GB consist of, because I run into external links in just a few clicks.

Any help would be greatly appreciated.

benoit74:
Fix #405 

Also contains a commit with test website changes for https://github.com/openzim/warc2zim/issues/403 (which might be only an upstream issue in the end, but always good to have test case on test website for that).

Also fixes type hints linked to recent python-libzim changes (the two later one are linked to https://github.com/openzim/python-scraperlib/issues/196 and will have to be reverted one this issue is fixed).


rgaudin (approved):
LGTM

PR #425, opened 2024-11-25
Stop checking main entry processability when it is already found

### LINKED ISSUE #424: thales_doc is considered unprocessable by mistake
Situation:
- we have one record for `https://thalesdoc.com/` which is OK (status code 200, nice content, ...)
- and we have one bad record for `https://thalesdoc.com//` which is KO (status code 403)

Both URL rewrite to the same ZIM Path `thalesdoc.com/` since we decide to ignore subsequent slashes: https://github.com/openzim/warc2zim/issues/316

This obviously has some adverse effects here. 

In general, having two entries leading to same ZIM Path, one with a bad HTTP return code and one with an OK one is not a big deal. One will simply be ignored, and the other added to the ZIM. 

But not when it comes to searching for main entry. I suggest we simply fix this "rare edge case" by stopping to search for main entry once it has been found. It is already the case in terms of main entry metadata - title, ... but not when it comes to main entry processability.

benoit74:
Fix #424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Upgrade to wombat 3.8.6 (#334)
 - Fix wombat setup settings (especially `isSW`) (#293)
 
+### Fixed
+
+- Stop checking main entry processability when it is already found (#424)
+
 ## [2.1.3] - 2024-11-01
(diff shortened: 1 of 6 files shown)


rgaudin:
Forgot to request review?

benoit74:
Really ready this time^^

rgaudin (approved):
LGTM

PR #446, opened 2025-03-17
provide default encoding aliases

### LINKED ISSUE #412: `--encoding-aliases` should be case insensitive
The aliases given in `--encoding-aliases` are currently case-sensitive. They should be case-insensitive (i.e. match even if case does not match).

### LINKED ISSUE #416: Provide default encoding aliases
The scraper would benefit a lot from having default encoding aliases already embedded. In many situations, we can be 99% sure of the proper encoding to use.

Defaults should be for now:
- `u` => `utf-8`
- `unicode` => `utf-8`
- `65001` => `utf-8`
- `urf-8` => `utf-8`
- `utf-f` => `utf-8`
- `utf-08` => `utf-8`
- `utp-8` => `utf-8`
- `windows-8859-1` => `iso-8859-1`
- `ansi` => `windows-1252`
- `uft-8` => `utf-8`
- `iso-utf-8` => `utf-8`
- `iso88591` => `iso-8859-1`

All these bad encoding have been found in the wild, and it is always a bit hard to fail a scrape "just for that", there is no ambiguities in these cases.

Setting the encoding aliases CLI arg would not drop these defaults, but complement/enrich them. The CLI value must however superseed the default (i.e. if `unicode` is redefined in `--encoding-aliases` CLI argument, it is the value from CLI - and not the default - which is used).

elfkuzco:
## Rationale
By providing a default map of "bad" encoding aliases to standard encoding names, the scraper can continue to decode the input without failing on encountering such aliases. This PR resolves #416 

Also, aliases provided via `--encoding-aliases` are converted to lowercase, thus making them case insensitive. This resolves #412 

## Changes
- Add a dictionary of default encoding aliases
- Provide method to retrieve an encoding method by alias as it simplifies testing without accessing the ENCODING_ALIASES directly.
- Make aliases to `--encoding-aliases` case insensitive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Provide default encoding aliases (#416)
+- Add `get_encoding_by_alias` method to retrieve an encoding by alias
+
 ## [2.2.2] - 2024-02-17
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
 - Provide default encoding aliases (#416)
-- Add `get_encoding_by_alias` method to retrieve an encoding by alias
+
+### Changed
+
+- Convert aliases given in `--encoding-aliases` to lower case (#412)
 
 ## [2.2.2] - 2024-02-17
 
--- a/src/warc2zim/utils.py
+++ b/src/warc2zim/utils.py
@@ -16,13 +16,32 @@ ENCODING_RE = re.compile(
     re.ASCII,
 )
 
+DEFAULT_ENCODING_ALIASES = {
+    "ansi": "windows-1252",
+    "65001": "utf-8",
+    "iso-utf-8": "utf-8",
+    "u": "utf-8",
+    "unicode": "utf-8",
+    "utf-8": "utf-8",
+    "utf-08": "utf-8",
+    "utf-f": "utf-8",
+    "utp-8": "utf-8",
+    "windows-8859-1": "iso-8859-1",
+    "iso88591": "iso-8859-1",
+}
+
 ENCODING_ALIASES = {}
 
 
 def set_encoding_aliases(aliases: dict[str, str]):
     """Set the encoding aliases to use to decode"""
     ENCODING_ALIASES.clear()
-    ENCODING_ALIASES.update(aliases)
+    ENCODING_ALIASES.update({**DEFAULT_ENCODING_ALIASES, **aliases})
+
+
+def get_encoding_by_alias(alias: str, default: str = "") -> str:
+    """Get the encoding method for alias."""
+    return ENCODING_ALIASES.get(alias, default)
 
 
 def get_version():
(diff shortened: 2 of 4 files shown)


src/warc2zim/utils.py
```diff
@@ -16,13 +16,32 @@
...
 
 
 def set_encoding_aliases(aliases: dict[str, str]):
     """Set the encoding aliases to use to decode"""
     ENCODING_ALIASES.clear()
-    ENCODING_ALIASES.update(aliases)
+    ENCODING_ALIASES.update({**DEFAULT_ENCODING_ALIASES, **aliases})
+
+
+def get_encoding_by_alias(alias: str, default: str = "") -> str:
```
benoit74, line 42: `alias` value should be lower-cased as well to match dictionary key even if case is wrong

benoit74: And since alias is always equal to default in our usage AFAIK, we should have only one parameter, this would be less confusing

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Provide default encoding aliases (#416)
+- Add `get_encoding_by_alias` method to retrieve an encoding by alias
```
benoit74, line 13: No need to specify this kind of changes (new methods/classes/..., any code-only changes) in the CHANGELOG, it does not matter at all to scraper users (scrapers are not a library with an API which can be reused by users, at least nobody does it AFAIK). We specify only behavioral / functional changes in scraper CHANGELOG

benoit74 (changes requested):
Thank you. Some changes needed.

CHANGELOG.md
benoit74: Sorry, I meant only that `Add get_encoding_by_alias method to retrieve an encoding by alias` change is not necessary. Other are required.

benoit74 (changes requested):
We misunderstood ourselves regarding the CHANGELOG + few other changes needed

src/warc2zim/utils.py
```diff
@@ -16,13 +16,32 @@
...
 def set_encoding_aliases(aliases: dict[str, str]):
     """Set the encoding aliases to use to decode"""
     ENCODING_ALIASES.clear()
-    ENCODING_ALIASES.update(aliases)
+    ENCODING_ALIASES.update({**DEFAULT_ENCODING_ALIASES, **aliases})
+
+
+def get_encoding_by_alias(alias: str) -> str:
+    """Get the encoding method for alias."""
+    return ENCODING_ALIASES.get(alias.lower().strip(), "")
```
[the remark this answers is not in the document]
elfkuzco: Okay. Do you mean using alias as default too?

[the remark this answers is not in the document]
benoit74: Yes, so that we keep same behavior as before, or do I miss something?

elfkuzco:
> We misunderstood ourselves regarding the CHANGELOG + few other changes needed

I have applied the changes and modified the CHANGELOG

tests/test_utils.py
```diff
@@ -361,3 +361,28 @@ def test_decode_charset_too_far_away_without_proper_alias():
             ignore_http_header_charsets=False,
             ignore_content_header_charsets=False,
         )
+
+
+@pytest.mark.parametrize(
```
benoit74, line 366: Please add those tests as well:
- `"   uNiCoDe "` => `"utf-8"`
- if you set `set_encoding_aliases({"uniCodE": "laTin1"})`, then `assert get_encoding_by_alias("unicode") == "latin1"` is still true and `assert get_encoding_by_alias("Unicode") == "latin1"` as well

benoit74: And add `assert get_encoding_by_alias("unKnoWn") == "unknown"`

elfkuzco: Um, the `set_encoding_aliases` never actually does any cleaning as that is done while parsing the input from the flag. If the tests involving `set_encoding_aliases` were to pass, I would have to move that logic to strip the input and lowercase it to the method rather than from the input flag. What do you think?

elfkuzco: Should I add this as a docstrinng instead given this method is never called except from the main? The docstring would explain that the keys and values should be lower-cased.

benoit74: You're right, then just check that `assert get_encoding_by_alias("Unicode") == "latin1"` is True in existing test case

elfkuzco: Already done.

benoit74 (approved):
LGTM, thank you!

